PR #1, opened 2020-09-06
First version of the Polylith tool

tengstrand:
Includes the last four months work with the new Polylith tool that is based on tools.deps.


furkan3ayraktar (approved):
Great job with this huge PR @tengstrand! 💯

PR #22, opened 2020-09-27
Update doc 11

tengstrand:
* Updated the doc: Context (new), Naming (new), Mix Languages (new) + more.
* Set correct :deps/root in deps.edn when creating a workspace. 
* Changed message to "Running tests from...".


tengstrand:
* Updated the doc: Context (new), Naming (new), Mix Languages (new) + more.
* Set correct :deps/root in deps.edn when creating a workspace.
* Changed message to "Running tests from...".

PR #62, opened 2020-12-02
Fix: Project Tests

tengstrand:
Calculate whether project tests should be executed based on affected affected projects instead of changed projects.


furkan3ayraktar:
`poly test :project` works as it should be now. However, neither `poly info` nor `poly info :project` command shows an asterisk and an `x` on the last column for the affected project. I think we should fix that also.

PR #64, opened 2020-12-19
Changed how to retrieve git tags

tengstrand:
Retrieve tags based on git logs instead of git tags, when finding the latest stable or release tag.

### DIFF
--- a/components/help/src/polylith/clj/core/help/diff.clj
+++ b/components/help/src/polylith/clj/core/help/diff.clj
@@ -15,7 +15,7 @@
        "\n"
        "  The pattern can be changed in " (color/purple color-mode ":stable-tag-pattern") " in ./deps.edn.\n"
        "\n"
-       "  The way the latest tag is found is by taking the first line that matches the 'table-*'\n"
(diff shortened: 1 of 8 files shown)


components/help/src/polylith/clj/core/help/diff.clj
```diff
@@ -15,8 +15,9 @@
        "\n"
        "  The pattern can be changed in " (color/purple color-mode ":stable-tag-pattern") " in ./deps.edn.\n"
        "\n"
-       "  The way the latest tag is found is by taking the last line of the output from:\n"
-       "    git tag --sort=committerdate -l 'stable-*'\n"
+       "  The way the latest tag is found is by taking the first line that matches the 'table-*'\n"
```
furkan3ayraktar, line 18: There is a typo here, instead of `table-*` it should be `stable-*`:
```suggestion
       "  The way the latest tag is found is by taking the first line that matches the 'stable-*'\n"
```

furkan3ayraktar (changes requested):

Issue #68, opened 2021-02-09
Suggestion: create component test ns without require


DavidVujic:
**Is your feature request related to a problem? Please describe.**
When creating a component, a test namespace is also added. That is nice, but it causes warnings when linting code.

**Describe the solution you'd like**
I suggest that it is added without the require statement, or maybe with a boilerplate unit test.

This is because when using tools like `clj-kondo` you will get lint warnings for unused requires.
I think it is this row that does it: https://github.com/polyfy/polylith/blob/master/components/creator/src/polylith/clj/core/creator/brick.clj#L32

PR #70, opened 2021-02-10
Remove issue with clj-kondo by adding dummy-test

tengstrand:
The [issue 68](https://github.com/polyfy/polylith/issues/68) suggests that we remove the :require statement from the generated test namespace.
Instead we now add a dummy test, to get rid of the clj-kondo problem, and instead keep the :require statement, to make it easier for the user to start writing tests.


tengstrand:
Generate dummy-test so that clj-kondo doesn't complain and that we can keep the :require statement.

PR #71, opened 2021-02-13
keep dependencies exclusions in test runner

mraveloarinjaka:
Hello,
In the test runner dependencies exclusions are lost when converting back to tools.deps format the project read from disk.
The aim of this fix is to make sure that they are kept.
Thanks


mraveloarinjaka:
Hello,
Note that I am not sure whether a pull request is the correct workflow you would like to have in place to contribute to polylith.
I can open an issue or you can have a look at the patch (it is a really small fix) and decide to integrate it directly.
Thank you.

tengstrand:
Thank you very much for this fix! 
You are very welcome to contribute by creating patches.

Issue #66, opened 2021-01-19
Use of :paths outside a project is not recommended


puredanger:
**Describe the bug**
The polylith directory structure instructs users to create projects with a deps.edn with `:paths` outside the project itself:

```clojure
:paths ["../../components/user/src"
         "../../components/user/resources"
         "../../bases/cli/src"
         "../../bases/cli/resources"]
```

This is generally a bad idea. The intent of `:paths` is to describe paths in your own project. In the polylith setup, these are effectively local dependencies. In the future, use of paths that are not under the project root may emit a warning or even an error.

deps.edn supports local dependencies using the `:local/root` coordinate attribute. Relative paths are supported there.

```clojure
:deps {my.component/user {:local/root "../../components/user"}
       my.base/cli {:local/root "../../bases/cli"}
```

You would also need deps.edn files under ../../components/user with `:paths ["src" "resources"]` etc of course.

I can't fully assess what other impacts this may have on the polylith architecture, but using local deps is strongly preferred over referring to non-project paths. If for some reason this is not sufficient, I'd appreciate talking through why (I know there are some issues others have raised for monorepo style setups).

Alternately, using a deps.edn at the root of the entire project may also be an option, if you want to consider the entire thing as a single project.

furkan3ayraktar:
Hi Alex! Thank you for your interest in Polylith and creating this issue to help us make it even better.

I’ll begin my response by quickly explaining the relevant parts of the high-level design of Polylith, and how that affects the various approaches we tried with `deps.edn`, before landing on using the relative paths.

Projects in a Polylith workspace are the way to configure deployable artifacts. Each project can be deployed as a jar file or as a Datomic Ions revision, etc. So the projects `deps.edn` is just a way to pick a set of components, a base, and external dependencies to create an artifact. Additionally, there is a `:test` alias in the `deps.edn` where the directories containing tests for each component are included.

Components are encapsulated blocks of code, which can be assembled together in projects with other components, a base, and external dependencies. We do not see them as local libraries or dependencies, rather we see them as a living part of a single codebase.

So that’s a quick recap of relevant parts of the [documentation](https://polylith.gitbook.io/polylith/), and now I can explain what we’ve tried so far.

[...]

PR #72, opened 2021-02-18
Expose interface file paths when reading the workspace

SneakyPeet:
### What?
Expose paths to interface files as part of the workspace.

`polylith.clj.core.workspace-clj.interface/workspace-from-disk` return data will now include a `:paths` field in the component interface data.

```clojure
{:components 
 [{:interface 
   {:paths ["/full/path/to/interface.clj"]}}]}
```

`polylith.clj.core.workspace.interface/enrich-workspace` will add the same `:paths` field to the interface data

```clojure
{:interfaces
 [{:path ["/full/path/to/interface.clj"]}]}
```

### Why?
It will be handy to be able to generate documentation for the interfaces of a given workspace. I looked into adding metadata, doc-strings and attr-map as part of the interface definitions. 

However due to the way poly tool currently reads files (using `clojure.core/read`) any metadata literals (`^{}`) defined does not get read and is not available for export. There are ways around this, but I doubt that it makes sense going through the effort to change how poly reads files.

Instead, if the paths to interfaces files are available on workspace level, anyone can use these paths as they see fit and integration with existing documentation tools become easier. 

The code change footprint is also much smaller

Finally the workspace file is less bloated than it would be if metadata is added (or alternatively if extra code has to be written to toggle metadata on and off).

### DIFF
--- a/components/workspace-clj/src/polylith/clj/core/workspace_clj/components_from_disk.clj
+++ b/components/workspace-clj/src/polylith/clj/core/workspace_clj/components_from_disk.clj
@@ -13,14 +13,16 @@
         src-dir (str component-src-dir interface-path-name)
         namespaces-src (ns-from-disk/namespaces-from-disk component-src-dir)
         namespaces-test (ns-from-disk/namespaces-from-disk component-test-dir)
-        definitions (defs-from-disk/defs-from-disk src-dir interface-ns)]
+        definitions (defs-from-disk/defs-from-disk src-dir interface-ns)
+        interface-paths (defs-from-disk/interface-paths src-dir interface-ns :include-root? true)]
     (util/ordered-map :name component-name
                       :type "component"
                       :namespaces-src namespaces-src
                       :namespaces-test namespaces-test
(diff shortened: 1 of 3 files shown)


SneakyPeet:
@tengstrand I've been mulling over this solution and I realise that it is just as easy to build the path from the actual ws data.

Also given #66 and the potential breaking changes, I'm not sure that this actually adds value (given that you can just build the paths).

So given the above I am going to close this PR for now. We can probably revisit it, but ultimately, sometimes the best code is no code :)

tengstrand:
@SneakyPeet My suggested solution is this:
- Read metadata and add the :meta key for all definitions in the interface if :meta is passed in as an argument (e.g. "components:mycomponent:interface:definitions"). 
- Add the attributes :is-interface and :sub-ns (if is an interface and a sub namespace) to :namespaces-src for each component (e.g. "components:mycomponent:namespaces-src").

This should make it easy to connect the paths to each definition, and also be a convenient way to extract metadata.

Issue #73, opened 2021-03-11
Poly command fails when source has non-seq import items


futuro:
**Describe the bug**
Having an import that isn't in a seq causes polylith cli to throw an exception. If you wrap your import in a vector or list, poly works as expected. This is also true for requires.

**To Reproduce**
1. Have a namespace with a bare import, like so
```clojure
(ns example.ns
  (:import java.time.LocalDateTime))
```
2. Run `poly info`
3. See error `Don't know how to create ISeq from: clojure.lang.Symbol`

**Expected behavior**
To see the poly info output, and not an exception string.

**Screenshots**
N/A

**Workspace Attachment**
I'm pretty sure this isn't applicable, but let me know if you need it.

**Operating System (please complete the following information):**
 - OS: Arch Linux
 - Version: Latest

**Versions (please complete the following information):**
 - Java: 15.0.2
 - Poly: Latest release, as well as hash `ec38a83db7597e6f6f33dfbc1e4da665807ab091`

**Additional context**

Stacktrace:
```
java.lang.IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.Symbol
 at clojure.lang.RT.seqFrom (RT.java:557)
    clojure.lang.RT.seq (RT.java:537)
    clojure.lang.RT.first (RT.java:693)
    clojure.core$first__5402.invokeStatic (core.clj:55)
    clojure.core/first (core.clj:55)
    polylith.clj.core.workspace_clj.namespaces_from_disk$import$fn__6076.invoke (namespaces_from_disk.clj:15)
    clojure.core$map$fn__5885.invoke (core.clj:2759)
    clojure.lang.LazySeq.sval (LazySeq.java:42)
    clojure.lang.LazySeq.seq (LazySeq.java:58)
    clojure.lang.Cons.next (Cons.java:39)
    clojure.lang.RT.length (RT.java:1785)
    clojure.lang.RT.seqToArray (RT.java:1726)
    clojure.lang.LazySeq.toArray (LazySeq.java:132)
    clojure.lang.RT.toArray (RT.java:1699)
    clojure.core$to_array.invokeStatic (core.clj:346)
    clojure.core$sort.invokeStatic (core.clj:3101)
    clojure.core$sort.invokeStatic (core.clj:3090)
    clojure.core$sort.invoke (core.clj:3090)
    polylith.clj.core.workspace_clj.namespaces_from_disk$imports.invokeStatic (namespaces_from_disk.clj:18)
[...]

PR #74, opened 2021-03-11
Handle imports that aren't seqs

futuro:
The current implementation assumes that each of the items in your
`:require` and `:import` forms are seqable, but that may not be the case,
at which point the poly_cli will throw an exception and quit.

This commit works around that by checking explicitly for whether the
import is seqable, and doing the right thing for each case.

As I'm writing some new tests, I've realized that my initial patch doesn't 
/quite/ work, because the expectation with imports, currently, is that
something like `[java.io File]` turns into `"java.io"`, so a naive call to `str`
doesn't return the same results.

This doesn't apply to a `:require` form, however, so that requires a bit
more finesse.

I'll work on a solution and push the changes.

See #73 for more details.

### DIFF
--- a/components/workspace-clj/src/polylith/clj/core/workspace_clj/namespaces_from_disk.clj
+++ b/components/workspace-clj/src/polylith/clj/core/workspace_clj/namespaces_from_disk.clj
@@ -5,11 +5,6 @@
             [polylith.clj.core.util.interface.str :as str-util])
   (:refer-clojure :exclude [import require]))
 
-(defn import? [statement]
-  (and
-    (sequential? statement)
-    (contains? #{:import :require} (first statement))))
-
 ;; (:require ,,,) handling
 
 (defn libspec?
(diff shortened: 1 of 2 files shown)


tengstrand:
Hi there!

We appreciate your effort, but a fix will soon be irrelevant because the tool will instead 
read what libraries each brick uses by looking in its `deps.edn` instead if parsing the code.

See my answer in [issue 73](https://github.com/polyfy/polylith/issues/73).

/Joakim

tengstrand:
Hi again!

I was a little bit too fast in my previous reply, which was more related to the :ns-to-lib mapping,
so my new answer to you is to continue with what you have started!

Thanks a lot,
/Joakim

tengstrand:
Hi again!

I would say option 1.

Even though it's not the recommended way, it's supported by the language
and it will not complicate the implementation + we don't need a warning.

futuro:
Hi @tengstrand!

I finally had a chance to sit down and write this up, and it's ready for review.

Thanks!
Evan

tengstrand:
Good job @futuro, looks great!

PR #75, opened 2021-03-27
fix misplaced docstrings

laurio:
fixed clj-kondo warnings about the misplaced docstrings


tengstrand:
Thanks for fixing this @laurio :-)

PR #77, opened 2021-03-28
Add calva jack in settings

PEZ:
Adding some Calva settings to the created workspace. (I'm still undecided about wether this is a good idea, but I lean towards that it is 😄).

Also noted that the created `.gitignore` lacked some common Clojure stuff and also some polylith stuff, so added most of the stuff from this project.


PEZ:
@seancorfield that is quite interesting. @furkan3ayraktar wondered the same thing with [my PR towards the RealWorld example.](https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/pull/15#pullrequestreview-622743276).  Same answer from me here; I think it might become a place where `clojure-lsp` wants to place config files. Maybe we should ping @ericdallo in to see if that is something to consider or not.

ericdallo:
Hey 👋🏻 
`clojure-lsp` ATM only creates the `.lsp/sqlite.db` with the analysis cache for faster next startups. it also reads a config file that user should create in its project: `project/.lsp/config.edn` or in the home: `~/.lsp/config.edn`.
More info here: https://clojure-lsp.github.io/clojure-lsp/settings/

seancorfield:
So it sounds like only the `sqlite.db` should be excluded from `git` @ericdallo ? And `.lsp/config.edn` _should_ be committed? (If so, I'll update `clj-new` to not ignore the whole folder)

ericdallo:
Yes @seancorfield, with that projects can have their own .lsp/config.edn for custom configuration

seancorfield:
@ericdallo Thanks. I've updated `clj-new` on *develop* and will note the change in the next release.

PR #79, opened 2021-04-13
Update uberdeps to be qualified

jacekschae:
When executing `clj -A:uberjar` commands the first thing we get is:
```DEPRECATED: Libs must be qualified, change uberdeps => uberdeps/uberdeps (deps.edn)``` 

PS. Not sure why this `@@ -2378,15 +2378,15 @@ Settings that are unique per developer/user are stored in ~/.polylith/config.ed` also got an edit?


tengstrand:
Looks good!

PR #81, opened 2021-04-29
Clarify development directory description

bpringe:
Hi! I just want to first say thank you for the time you've put sharing this tool and the Polylith architecture. The gitbook is very polished and concise, and this readme looks very thorough. I've read through the docs and watched some videos and am now trying out the walkthrough in this readme.

In the part describing the directories created by the tool, "development project" seemed a bit confusing to me at first glance, and I looked into the project directory just to make sure there wasn't some "project" called development as well. I'm not sure if this change makes sense to you, but if so maybe it can be merged.


bpringe:
Nevermind, I think I misunderstood. Noob mistake!

PR #82, opened 2021-04-30
Fully qualify uberdeps dependency

bpringe:
I'm not sure if this is something that's going to be enforced by tools.deps in the future, but I think it's recommended to fully qualify deps in this way.

clj-kondo warns of this when it's just `uberdeps`: `Libs must be qualified, change uberdeps => uberdeps/uberdeps`


tengstrand:
Yes, they should be fully qualified. Thanks!

PR #95, opened 2021-08-02
Use depstar for deployer

furkan3ayraktar:
- Use [seancorfield/depstar](https://github.com/seancorfield/depstar) instead of [pack/pack.alpha](https://github.com/juxt/pack.alpha) in the `deployer` component for deploying Polylith artifacts to [clojars](https://clojars.org).
- Fixed invalid usage of [slipset/deps-deploy](https://github.com/slipset/deps-deploy). The version of the library is incremented during the development of #94 but the way we use was not updated accordingly. Now that is fixed.


tengstrand (approved):
Nice work!

Issue #97, opened 2021-08-03
Remove validation that stops a workspace to be created within another workspace


tengstrand:
Allow a workspace to be created in existing workspaces by deleting the old validation "A workspace should not be created within another workspace.".

PR #98, opened 2021-08-03
Remove validation that stops you from creating a workspace within another workspace

tengstrand:
Solves issue #97


furkan3ayraktar (approved):
💪

PR #111, opened 2021-08-13
Fix #108 by respecting deps/root in git deps

### LINKED ISSUE #108: Test Suite Cannot Resolve Multiply-Required Transitive Dependency
**Describe the bug**

`poly test` cannot run if two components have transitive dependecies on the same Github-hosted library.

**To Reproduce**
1. Clone a repo [I made to exhibit this behavior.](https://github.com/Matthew-Ielusic/MinimalBugRepro)
2. Run `poly test`
3. See error:
```Couldn't resolve libraries for the deploy project: clojure.lang.ExceptionInfo: Unable to compare versions for repro-lib/root: {:local/root "/Users/<etc.>```

**Expected behavior**
The test suite should run without issue - in fact, it does run without issue in Kaocha, through the script `kaocha.sh`.

**Operating System (please complete the following information):**
 - OS: MacOS
 - Version: 11.4 (20F71)

**Versions (please complete the following information):**
 - Java: 
```
openjdk 11.0.11 2021-04-20
OpenJDK Runtime Environment AdoptOpenJDK-11.0.11+9 (build 11.0.11+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK-11.0.11+9 (build 11.0.11+9, mixed mode)
```

 - Poly: 0.2.0-alpha11
 - Brew: 
```
Homebrew 3.2.6
Homebrew/homebrew-core (git revision fb7013ba7a; last commit 2021-08-11)
Homebrew/homebrew-cask (git revision d863760d7b; last commit 2021-08-10)
```

**Additional context**

To be clear:

- The repo I linked to has two components, `left` and `right`
- Those components depend on two different libraries [hosted in another repo I made](https://github.com/Matthew-Ielusic/ReproLibrary)
- Those two libraries each depend on a third library, hosted in the same repo
- I added that shared library to `:exclusions` in all the deps.edn files.

### PR DESCRIPTION
**Pull request by seancorfield**
Also supports old and new forms of tag and sha.
Also supports exclusions for ALL dep types.

### DIFF
--- a/components/deps/src/polylith/clj/core/deps/lib_deps.clj
+++ b/components/deps/src/polylith/clj/core/deps/lib_deps.clj
@@ -4,13 +4,27 @@
 (defn name-version [[k {:keys [mvn/version]}]]
   [(str k) version])
 
-(defn adjust-key [{:keys [type path version git/url sha exclusions]}]
-  (case type
-    "maven" {:mvn/version version
-             :exclusions (vec exclusions)}
-    "local" {:local/root path}
-    "git"   {:git/url url :sha sha}
-    (throw (Exception. (str "Unknown library type: " type)))))
+(defn adjust-key
+  "git deps can have sha or git/sha, tag or git/tag, and deps/root
+  These days, git/url can be omitted and deduced from the lib name.
+
+  All deps can have exclusions."
+  [{:keys [type path version
+           git/url git/sha git/tag deps/root
+           exclusions]
+    old-sha :sha old-tag :tag}]
+  (cond->
+   (case type
+     "maven" {:mvn/version version}
+     "local" {:local/root path}
+     "git"   (cond-> {}
+               url              (assoc :git/url url)
+               (or sha old-sha) (assoc :git/sha (or sha old-sha))
+               (or tag old-tag) (assoc :git/tag (or tag old-tag))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: components/deps/src/polylith/clj/core/deps/lib_deps.clj [hunk 1]
```diff
@@ -4,13 +4,27 @@
...
+           exclusions]
+    old-sha :sha old-tag :tag}]
+  (cond->
+   (case type
+     "maven" {:mvn/version version}
+     "local" {:local/root path}
+     "git"   (cond-> {}
+               url              (assoc :git/url url)
+               (or sha old-sha) (assoc :git/sha (or sha old-sha))
+               (or tag old-tag) (assoc :git/tag (or tag old-tag))
```
imrekoszo, line 23: Might be a good idea to also assoc the old style sha and tag for better backward compatibility with older cli tools?

seancorfield: `poly` uses the latest `tools.deps.alpha` and this is for its own use so it's just making them canonical internally. The idea here is that regardless of which git deps format you use in `deps.edn` in a Polylith project, it will understand it and produce a standard hash map (for later consumption by t.d.a).

tengstrand (approved):
Looks good!

PR #115, opened 2021-08-21
Add Clojure CLI API for -X/-T

seancorfield:
I didn't realize you have a lot of trailing whitespace in the `readme.md` file and my editor is set to automatically trim that. If you want me to redo just the `readme.md` without trimming the whitespace, I can probably do that. It's just one new section (at the end of the current install section, before the section on using Polylith as a dependency):

### Install the Polylith Tool via the Clojure CLI

If you are using a recent version of the [Clojure CLI](https://clojure.org/releases/tools) --
1.10.3.933 or later -- you can install `poly` as a "tool":

```sh
clojure -Ttools install io.github.polyfy/polylith '{:git/sha "..." :deps/root "projects/poly"}' :as poly
```

**TODO: This needs updating with the latest full SHA after merging the PR, and then with `:git/tag`/`:git/sha` (using a short SHA) after the need tagged version is released!**

Then you can invoke the `poly` tool using the CLI directly:

```sh
clojure -Tpoly info :loc true
```

You can get basic built-in help via the CLI's help machinery:

```sh
clojure -A:deps -Tpoly help/doc
```

> Note: the command-line argument syntax for "tool" usage follows the CLI "exec args" format which is essentially Clojure's own keyword/value syntax (it's actually read as EDN).

#### Exec Arguments vs Polylith Tool Documentation

The "native" `poly` tool has the following argument structure:
* the command, followed by
* zero or more entities (e.g., `create` can be followed by `component`), followed by
* named arguments (e.g., `project:example`), and
* flag arguments (e.g., `:loc`), and
* profiles (e.g., `+remote`).

Because "exec args" must follow keyword/value syntax, the argument structure for `-Tpoly` is:
* the command (which is the "exec fn"), followed by
* `:entity` to name the first argument(s) to `create` or `help`,
* named arguments as keyword/value pairs (e.g., `:project example`),
* flag arguments as keywords with `true` or `false` values (e.g., `:loc true`),
* `:profiles` to list any profiles to enable, without the leading `+` (e.g., `:profiles '[remote]'`).

For convenience, most arguments can be provided as symbols rather than requiring quoted strings:

```sh
# strings have to be quoted for EDN and the shell:
clojure -Tpoly info :project '"example"'
# the following shorthand is allowed:
clojure -Tpoly info :project example
```

Both `:project` and `:profile` exist in singular form, taking a symbol or string, and in
a plural form, taking a vector of symbols or strings:

```sh
# select more than one project:
clojure -Tpoly info :projects '[example cli]'
# select just one profile:
clojure -Tpoly info :profile remote
```

Some `poly` commands treat `:brick` and `:project` as flag arguments and other expect
`brick:name` or `project:name` as named arguments. When using "exec args", the meaning
is determined by whether the value is boolean or not:

```sh
# select all projects:
clojure -Tpoly info :project true # flag argument
# select a specific project:
clojure -Tpoly info :project cli # named argument
```

Since the exec argument format for create becomes verbose, there are four shortcut arguments:

```sh
# full create command syntax:
clojure -Tpoly create :entity component :name user
# shorthand:
clojure -Tpoly create :c user
# similarly for :b (base), :p (project), and :w (workspace)
clojure -Tpoly create :w next-gen :top-ns com.my-company
# is shorthand for:
clojure -Tpoly create workspace :name next-gen :top-ns com.my-company
```

Finally, for some commands, more than one "entity" can be provided and for the `:entity`
exec argument, those can be `:`-separated:

```sh
clojure -Tpoly help :entity create:component
```

### DIFF
--- a/bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
+++ b/bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
@@ -1,0 +1,138 @@
+(ns polylith.clj.core.poly-cli.api
+  "An exec fn API for the poly tool that can be used
+  with -X and -T options."
+  (:refer-clojure :exclude [test])
+  (:require [clojure.string :as str]
+            [polylith.clj.core.command.interface :as command]))
+
+(set! *warn-on-reflection* true)
+
+(defn- str-coll
+  "Given an argument value, coerce it to a vector and
+  make all the elements strings."
+  [arg]
+  (mapv str (if (vector? arg) arg [arg])))
+
+(defn- argument-mapping
+  "Map exec args to the sort of internal arguments that
+  the poly tool expects. This is made harder by the way
+  that poly has a somewhat 'unique' approach to command
+  line argument handling!
+
+  Keyword differences:
+  :brick    bool -- :is-show-brick true/false
+  :brick    s    -- :brick name
+  :entity   s    -- needed for the create and help commands
+  :profile  s    -- :selected-profiles (singular for convenience)
+  :profiles [s]  -- :selected-profiles (b/c no unnamed args)
+                    (may be just a single symbol/string)
+  :project  bool -- :is-show-project true/false
+  :project  s    -- :selected-projects (singular for convenience)
+  :projects [s]  -- :selected-projects (b/c avoid :-separated)
+                    (may be just a single symbol/string)
+  :ws       bool -- :is-search-for-ws-dir (b/c :: is not legal)
+
+  For create -- to make life easier:
+  :c s -- component name:s
+  :b s -- base name:s
+  :p s -- project name:s
+  :w s -- workspace name:s
+
+  Items not yet properly handled:
+  :args -- there's no concept of sequential ordered args
+    (a vector of the command + entity is passed)
+  :replace -- relies on non-interface function
+  :unnamed-args -- there's no concept of sequential ordered args
+    (a vector of the command + entity is passed)"
+  [exec-args]
+  (reduce-kv (fn [m k v]
+               (cond-> m
+                 ;; add is-* variant for all booleans:
+                 (boolean? v)
+                 (assoc (keyword (str "is-" (name k))) v)
+                 ;; add is-show-* variant:
+                 (and (boolean? v)
+                      (contains? #{:brick :workspace :project :loc :resources} k))
+                 (assoc (keyword (str "is-show-" (name k))) v)
+                 ;; some special case boolean flags:
+                 (and (boolean? v)
+                      (contains? #{:ws} k))
+                 (assoc :is-search-for-ws-dir v)
+                 (and (boolean? v)
+                      (contains? #{:all :all-bricks} k))
+                 (assoc :is-run-all-brick-tests v)
+                 (and (boolean? v)
+                      (contains? #{:all :project} k))
+                 (assoc :is-run-project-tests v)
+                 (and (boolean? v)
+                      (contains? #{:r} k))
+                 (assoc :is-show-resources v)
+                 ;; vectorize skip:
+                 (contains? #{:skip} k)
+                 ;; stringify any skipped projects:
+                 (assoc :skip (when v (str-coll v)))
+                 ;; profile(s):
+                 (contains? #{:profile} k)
+                 (update :selected-profiles conj (str v))
+                 (contains? #{:profiles} k)
+                 (update :selected-profiles into (str-coll v))
+                 ;; project(s) + dev (flag):
+                 (and (not (boolean? v))
+                      (contains? #{:project} k))
+                 (update :select-projects conj (str v))
+                 (contains? #{:projects} k)
+                 (update :selected-projects into (str-coll v))
+                 (contains? #{:dev} k)
+                 (cond-> v (update :selected-projects conj "dev"))
+                 ;; stringify any possible symbol args:
+                 ;; (this is just a usability issue for exec args)
+                 (and (not (boolean? v))
+                      (contains? #{:get :brick :branch :color-mode
+                                   :entity
+                                   :fake-sha :interface :name :out
+                                   :since :top-ns :ws-dir :ws-file} k))
+                 (update k str)
+                 ;; special handling for create shorthands:
+                 (contains? #{:c :b :p :w} k)
+                 (assoc :entity (name k) :name (str v))))
+             ;; defaults
+             (merge {:selected-profiles #{}
+                     :selected-projects #{}
+                     :unnamed-args []}
+                    exec-args)
+             exec-args))
+
+(defn- ->command
+  "Given a command name and the exec args, map those to
+  internal arguments, and execute the command."
+  [cmd exec-args]
+  (let [args     (argument-mapping exec-args)
+        ;; help is a special case for :entity as it can
+        ;; have multiple arguments, so we :-separate them
+        entities (when-let [entity (:entity args)]
+                   (str/split entity #":"))
+        args     (assoc args
+                        :args (cond-> [cmd]
+                                entities (into entities))
+                        :cmd  cmd)
+        exit
+        (command/execute-command args)]
+    ;; this mirrors the cli/-main behavior:
+    (when-not (:is-no-exit args)
+      (System/exit exit))))
+
+;; these are the exec fns that correspond to commands:
+
+(defn check   [exec-args] (->command "check"   exec-args))
+(defn create  [exec-args] (->command "create"  exec-args))
+(defn deps    [exec-args] (->command "deps"    exec-args))
+(defn diff    [exec-args] (->command "diff"    exec-args))
+(defn help    [exec-args] (->command "help"    exec-args))
+(defn info    [exec-args] (->command "info"    exec-args))
+(defn libs    [exec-args] (->command "libs"    exec-args))
+(defn migrate [exec-args] (->command "migrate" exec-args))
+;; shell -> prompt
+(defn shell   [exec-args] (->command "prompt"  exec-args))
+(defn test    [exec-args] (->command "test"    exec-args))
+(defn version [exec-args] (->command "version" exec-args))
+(defn ws      [exec-args] (->command "ws"      exec-args))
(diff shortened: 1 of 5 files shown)


bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
```diff
@@ -0,0 +1,284 @@
...
+  :w s -- workspace name:s
+
+  Items not yet properly handled:
+  :args -- there's no concept of sequential ordered args
+    (a vector of the command + entity is passed)
+  :replace -- relies on non-interface function
+  :unnamed-args -- there's no concept of sequential ordered args
+    (a vector of the command + entity is passed)"
+  [exec-args]
+  (reduce-kv (fn [m k v]
```
tengstrand: The drawback with populating the keywords directly, e.g. setting `:is-run-project-tests` to `true`, is that we bake in some business logic here. An alternative is to produce a vector of strings, one per argument, e.g. [":all"] instead of directly setting `:is-all` to true. Then this list could be sent to the same `params/extract` function that is already used by the `user-input` component itself, and we could also send the produced statement to the [portal](https://github.com/djblue/portal) window (yes, that will be included soon!) as a way of getting feedback when developing functionality like this.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
```diff
@@ -0,0 +1,284 @@
...
+(defn- argument-mapping
+  "Map exec args to the sort of internal arguments that
+  the poly tool expects. This is made harder by the way
+  that poly has a somewhat 'unique' approach to command
+  line argument handling!
+
+  Keyword differences:
+  :brick    bool -- :is-show-brick true/false
+  :brick    s    -- :brick name
+  :entity   s    -- needed for the create and help commands
```
tengstrand: An alternative would be to have four separate commands, `create-base`, `create-component`, `create-project` and `create-workspace`. They could maybe be used by the help command also, e.g. `clojure -Tpoly help :entity create-base`. Another reflection. Maybe we can skip the `:` prefix, and instead of typing `clojure -Tpoly help :entity create` we type `clojure -Tpoly help entity create`.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
seancorfield: That is not idiomatic for the CLI which is why I avoided it. The `poly` tool already has "unusual" semantics because of its custom argument parsing. If it used POSIX-style arguments, and `clojure.tools.cli`, it would be much easier to map to exec arguments.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
seancorfield: The CLI exec arguments approach would support symbol-based keys, it's true, but `poly` already has a weird mixture of keyword-like things (`:loc`, `:project`, `:brick`, etc) and custom semantics with `foo:bar` being interpreted as a key/value pair.

Besides, on this particular case, I did introduce four separate keywords to shorten the command-line:
```
clojure -Tpoly create :c user
```
Here, `create` is an exec fn and `:c` / `user` is a keyword/value pair for the argument, intended as a shortcut for:
```
clojure -Tpoly create :entity c :name user
```

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
tengstrand: I think you misunderstood what I meant. I meant how to best generate the internal `user-input` hash map, by using the `params/extract` function instead of creating it directly.

tengstrand: It may not be standard, but it turns out to work really well with the new `shell` command that I'm working on (that supports autocomplete). If you start typing `:` then it will suggest alla available "flags" for the current command. By grouping the values for a parameter, e.g. `project:a:b` or `get:settings:components:util`, you can then press `tab` on the keyboard to instruct the autocompleter to list all the available candidates (projects or keys in this case) and it will add it at the end of the current parameter when you select one. You can continue adding values for that parameter till you are satisfied and then move on to the next parameter by pressing `space`.

seancorfield: Ah, so you're suggesting that the API exec fn turn the exec args -- a hash map read from the command-line as EDN -- into a vector of `poly` arguments and then call `extract` to re-parse that back into the internal format? That would remove the issue of keeping the hash map in sync between the two "APIs". OK, I'll update the PR with that approach. It may, ultimately, be simpler.

seancorfield: Exec API updated to produce `poly` tool style arguments sequence and then it calls `user-input/extract-params` and `command/execute-command` -- exactly like the `-main` entry point. I could have called `-main` directly but wanted to keep the exec API separate so that if the Clojure CLI ever gets multi-function `-X` invocation back, we can change the exec API behavior to work with that (i.e., not calling `System/exit`, which is not really idiomatic for exec fns).

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
seancorfield: If you want the exec API to support symbol-based keys (as well as keyword-based keys), I can easily update the logic for that.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
```diff
@@ -0,0 +1,247 @@
...
+  clojure -Tpoly help :entity create:component
+  clojure -Tpoly help :entity create:project
+  clojure -Tpoly help :entity create:workspace"
+  [exec-args]
+  (->command "create"  exec-args))
+
+(defn deps
+  "Shows dependencies.
+
+  clojure -Tpoly deps [:project PROJECT] [:brick BRICK]
```
tengstrand: I would prefer to only support the syntax `project PROJECT` to mimic `project:PROJECT` (and `project [P1 P2 P3]` if more than one project?) and `project true` to mimic `:project`. It feels like the colons don't add any value in this context.

tengstrand (commented):
I made a small objection, but except from that it looks really good.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
tengstrand: Or maybe use the syntax `is-project true` for `:project`. If we try to use only `project` then things like `poly info project:dev :project` needs to be translated to `-Tpoly info project [dev true]` instead of `-Tpoly info project dev is-project true`.

tengstrand: Another alternative to `:project true` is to use `is-project true`. It feels more readable to me.

tengstrand: That will make some sense also. Today `project:dev` is equivalent to `:dev`. With the suggested format, then `project dev` would be equivalent to `is-dev true` which reads quite nice to me.

seancorfield: OK, PR updated to accept symbols as keys for exec args (it still accepts keywords too) and docs updated to indicate symbols as the preferred use:
```sh
clojure -Tpoly info deps project remote brick user
```
You can provide a `:`-separated list of items for things: `project remote:cli`
You can provide a vector: `project '[remote cli]'` (quotes needed on the command-line)

For `entity`, `profile`, and `project` there are plurals accepted too: `projects '[remote cli]'`

seancorfield: For flag arguments via exec args, I expect folks will instinctively use `:loc true` rather than `loc true` which is why I figured it might as well accept _both_ (and convert to `:loc` as a flag argument for the internal argument parsing).

Right now, any boolean gets turned into a flag -- so `dev false` would also become `:dev` since I didn't think folks would bother specifying a _negative_ flag, but it would be a very small change to make `dev false` be simply ignored. The docs only state that flag arguments need to be given as `<flag> true`.

seancorfield: For the `info project:dev :project` example, `info project dev project true` would work, as would `info dev true project true` (I think? `:dev` as a flag argument just causes `dev` to be added as a selected project, right?). It's an edge case that we can ignore I think -- both `:project` and `:brick` can be flag arguments or named arguments.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
seancorfield: The last update made symbols work, and that is what is documented now, although keywords can still be used. Let me know if there are any other issues to consider/clean up.

bases/poly-cli/src/polylith/clj/core/poly_cli/api.clj
tengstrand: Yes, `:dev` will add `development` to the list of projects. One idea I have is to support passing in all the arguments as plain "polylith style" text, like this: `clojure -A:deps -Tpoly 'help deps :brick'` or if the `-T` needs an extra argument, like this: `clojure -A:deps -Tpoly apply 'help deps :brick'`.

seancorfield: That's just not how the CLI works for tools. The arguments **must** be key/value pairs. To pass a string like that you'd have to do:

```sh
clojure -Tpoly apply :args '"help deps :brick"'
```
Note the two sets of quotes which are required because the key/value pairs must parse as EDN and each shell requires certain amounts of quoting. The above would work on macOS/Linux but not on Windows.

As the PR stands, the above command would be:
```sh
clojure -Tpoly help entity deps brick true
```
which would work on all platforms without quoting. `entity` and `brick` could be keywords if folks want.

tengstrand: My idea was to trick the CLI by saying that the command (first argument after `poly`) is `'"help deps :brick"'` which would allow people to write `clojure -Tpoly '"help deps :brick"'`, and we can use the string internally "as is". We will also support the CLI style format, as suggested in this PR, e.g. `clojure -Tpoly help entity deps brick true`.

seancorfield: That simply isn't possible. The "first argument after `poly`" is a Clojure function name -- it can't be an arbitrary string.

tengstrand (approved):
It looks good. Thanks for this @seancorfield.

Issue #110, opened 2021-08-12
Per-project test run setup and teardown


seancorfield:
**Is your feature request related to a problem? Please describe.**
At work, when we run our legacy test suite, we perform some overall setup and teardown around the actual test run that every test would need but would be time-consuming to setup and teardown around every single namespace (i.e., what would be the `:once` fixture from `clojure.test`). I think this is a fairly common need, and some test runners even provide these two additional hooks beyond what `clojure.test` itself provides.

**Describe the solution you'd like**
I'd like to be able to specify `:test-setup` and `:test-teardown` as optional keys in each project's configuration (that already includes `:alias` and `:test`) to provide fully-qualified symbol names, and then have `poly test` invoke these at the start of running tests for a project and at the end.

**Describe alternatives you've considered**
For Cognitect's `test-runner` in our legacy code setup, we have our own wrapper that we invoke, which in turn invokes `test-runner, and that wrapper is where we perform setup and teardown. This isn't possible with `poly test` so an alternative would be for use to add a `:once` fixture to _every single test namespace_ that used a fixture with some global state to track whether the per-suite setup has yet been performed -- but this wouldn't support teardown. We could add teardown to the JVM shutdown via a hook but that won't work when running `test` from the `poly shell`.

**Additional context**
Since `poly` runs tests in a classloader, the setup code would look something like this, added before the `doseq` over the `test-statements` (and similarly for `:test-teardown` after that `doseq`):

```clojure
(when-let [fixture (:test-setup project-map)]
  (try
    (common/eval-in class-loader
      (str "((requiring-resolve '" fixture "))"))
    (catch Throwable t
      (println "Test setup failed:" t))))
 ```
(perhaps with better error handling/messaging!)

tengstrand:
What you suggest looks great to me, and the issue is well written, so I think I have everything I need to implement this.
I will have a look at it as soon as I finish what I'm working on.

seancorfield:
I've run into a problem with this: as it stands, if the test suite has any failures, the `teardown-fn` is not executed -- it should always be executed.

PR #130, opened 2021-09-10
#110 should always run teardown-fn

seancorfield:
This fixes the problem I mentioned in #110 where a test failure prevents the teardown function from being executed.


tengstrand (approved):
Looks good!

PR #142, opened 2021-10-07
Provide tools.build-based deployment

seancorfield:
```bash
# build uberjar and homebrew stuff:
clojure -T:build create-artifacts
# build & deploy library JARs (locally):
clojure -T:build deploy :installer :local
# build & deploy library JARs (to Clojars):
clojure -T:build deploy
# get help on the build system:
clojure -A:deps -T:build help/doc
```

You will need at least 1.10.3.933 of the Clojure CLI for this (see https://clojure.org/releases/tools for latest versions).

The `build.sh` script now does `clojure -T:build uberjar :project poly` (and then `sudo cp ...`).

I have deliberately made the new `build` base & `build` project rely on as little of the Polylith code base as possible -- only the `version` component is needed. There's no real reason to use a base & project here, to be honest, the code could simply be `build.clj` in the workspace root, but it would still need a `:local/root` dep on `version` -- and I'd be happy to make that change if you think it would be less confusing than having the "build system" be part of the codebase.

I have not deleted anything (not `java`, not the `deployer`, not the `build*.sh` scripts).

I have (temporarily) restored the `<dependencies>` section to the `partial_pom.xml` files so that you can still run the `deployer` process to make it easier to compare results.

A "final" version of `tools.build`-based deployment should:
* remove the `java` folder,
* remove the `deployer-cli` and `deployer` bricks,
* remove the whole `<dependencies>` section from both `partial_pom.xml` files (some other parts could be removed too since nearly everything is automatically updated now).

Let me know if you have any questions or comments, or want to discuss changes.


furkan3ayraktar (approved):
Really great work @seancorfield ! I'm going to merge this branch and update the CircleCI related parts on my own branch. Thanks a lot again for the PR!

Issue #171, opened 2021-12-20
Make use of custom maven root in the test command.


tengstrand:
See Slack [thread](https://clojurians.slack.com/archives/C013B7MQHJQ/p1639654934403100?thread_ts=1639265474.384000&cid=C013B7MQHJQ)

tengstrand:
Fix merged to master.

PR #173, opened 2021-12-27
Use correct path to maven repository when running tests

tengstrand:
Add the configured m2-dir + `/repository` to the configuration (key `:mvn/local-repo`) when running tests, if it's configured as something else than `~/.m2` in `~/.polylith/config.edn`.


tengstrand:
Resolves issue #171

PR #178, opened 2022-02-01
Bugfix: Correct hash-sum denotation

TimoKramer:
The hash-sum is a sha256-sum not a sha1-sum. This leads to confusion.


tengstrand (approved):
Thanks for this patch!

PR #183, opened 2022-02-22
fix: calculation of file size for rare cases

jeroenvandijk:
`poly info` was failing for me because the poly command was walking over a directory with a `.#test.clj` file (emacs) which apparently was neither a file nor a directory. 

This commit adds a default `0` size for these exceptions to prevent a Nullpointer exception in the addition.


tengstrand:
Thanks for the PR!

PR #194, opened 2022-03-22
Fix some dysfunctional test assertions

imrekoszo:
Found these while hacking on the test runner.


tengstrand:
I have fixed these already, but thank for the PR!

PR #196, opened 2022-03-24
feat: pluggable test runner

imrekoszo:
Add the option to supply a custom test runner.

A previous version of this PR contained an example implementation using kaocha, I will recreate that outside of this repo and link it.

Todo:

- documentation
  - [x] document plugin and constructor interface
  - [x] document how to pass extra args to test runner using user-input (focus) and stuff like verbose
- [x] create repo for kaocha impl that was pulled from this PR
  - https://github.com/imrekoszo/polylith-kaocha
- [x] create issue about namespaces vs paths
- later perhaps
  - [ ] extend info integration for test discovery? potentially with a flag that hides a potentially slow `tests-present?` 
    - this is proving to be somewhat difficult as the source data for this gets collected rather early: https://github.com/polyfy/polylith/blob/e874e809aa9d3cecaa0540376bd2875f99b37372/components/workspace/src/polylith/clj/core/workspace/project.clj#L40-L50


components/kaocha-test-runner-example/src/polylith/clj/core/kaocha_test_runner_example/core.clj
```diff
@@ -0,0 +1,88 @@
...
+                                                  #_{:kaocha.testable/type :kaocha.type/spec.test.check
+                                                     :kaocha.testable/id :generative-fdef-checks}]}))))]
+    (reify test-runner-plugin/TestRunner
+      (test-sources-present? [_] test-sources-present)
+
+      (tests-present? [this {:keys [eval-in-project] :as _opts}]
+       (and (test-runner-plugin/test-sources-present? this)
+            (let [config (->test-plan-config eval-in-project)]
+              (eval-in-project
+               `(do (require ~''kaocha.api ~''kaocha.testable ~''kaocha.hierarchy)
```
kkrausse: I don't think the `~'` is required here. And same with below. I don't think syntax quote auto-namespaces symbols with dots (or slashes) in them

imrekoszo: You're right, I'm removing them.

components/clojure-test-test-runner/src/polylith/clj/core/clojure_test_test_runner/core.clj
```diff
@@ -0,0 +1,97 @@
...
+        (throw (Exception. (str "\n" (color/error color-mode result-str)))))
+      (println (str "\n" (color/ok color-mode result-str))))))
+
+(defn create
+  [{:keys [workspace project changes #_test-settings]}]
+  (let [{:keys [bases components]} workspace
+        {:keys [name namespaces paths]} project
+        {:keys [project-to-bricks-to-test project-to-projects-to-test]} changes
+
+        ;; TODO: if the project tests aren't to be run, we might further narrow this down
```
furkan3ayraktar, line 76: Is this something we will address now or a TODO for after we merge this PR?

components/common/src/polylith/clj/core/common/class_loader.clj
```diff
@@ -72,5 +72,6 @@
       (eval-in* class-loader '(require 'clojure.main))
       class-loader)
     (catch Exception e
+      ;; TODO: this seems to not print an error when clojure.lang.RT couldn't load
```
furkan3ayraktar: Hmm, two questions here:
1. In which case `clojure.lang.RT` could not load?
2. Could we print an error in a different way here so that we do not loose the error message? If we can't maybe we can just turn this into a comment rather than a TODO

components/workspace-clj/src/polylith/clj/core/workspace_clj/project_settings.clj
```diff
@@ -1,12 +1,24 @@
...
+(defn convert-test-fn [global-test]
+  (fn convert-test [[k {:keys [test] :as v}]]
+    [k
+     (cond-> v
+       (vector? test) (assoc :test {:include test})
+       global-test (update :test #(merge global-test %)))]))
+
+(defn or-default-constructor [candidate]
+  (if (contains? #{nil :default} candidate)
+    'polylith.clj.core.clojure-test-test-runner.interface/create
```
furkan3ayraktar, line 12: Since this is a symbol, in case of a refactoring we could easily forget this. Also Polylith will not understand that `workspace-clj` actually depends on `clojure-test-test-runner`.

How could we overcome this? Could we maybe just require `clojure-test-test-runner` in this namespace?

components/clojure-test-test-runner/src/polylith/clj/core/clojure_test_test_runner/core.clj
imrekoszo: I intended this as a note for later as I don't want to change what the default test runner does as part of this effort.

components/common/src/polylith/clj/core/common/class_loader.clj
imrekoszo: I found this while iterating on an early version of the kaocha runner implementation and had an error in the kaocha-specific code which manifested as the load error. What I noticed that the error wasn't getting printed.

Let me try and set up a repro case somewhere.

components/workspace-clj/src/polylith/clj/core/workspace_clj/project_settings.clj
imrekoszo: We could require for sure, but that would mean it will load that namespace even when someone does not want to use `clojure-test-test-runner`. As long as the polylith workspace uses the default test runner assigned here, `poly check` and `poly test` will both fail if this gets outdated.

components/common/src/polylith/clj/core/common/class_loader.clj
imrekoszo: Haven't managed to repro this, it was probably not a good idea to not commit a failing setup when I came across it.

In any case, the error that happened here was caused by me setting up something wrong. For now I removed this comment and will raise an issue with a repro case if I come across it again.

components/clojure-test-test-runner/src/polylith/clj/core/clojure_test_test_runner/core.clj
furkan3ayraktar: Alright, then we can leave this TODO for later 👍

components/common/src/polylith/clj/core/common/class_loader.clj
furkan3ayraktar: Sounds good

components/workspace-clj/src/polylith/clj/core/workspace_clj/project_settings.clj
furkan3ayraktar: Alright, based on our discussion on Slack, this should be okay to be left as it is and an explicit test could be added to the `poly` project.

imrekoszo: Added in https://github.com/polyfy/polylith/pull/196/commits/68af10c8e144901e67dc861bd571db6cc32306f1

SavageMessiah:
Thank you for working on this, I've just started porting our 150k line codebase over to polylith and management has been breathing down my neck about code coverage stats 😓 .

imrekoszo:
> coverage stats 😓 .

Glad someone else finds this useful :) I haven't yet set up coverage on our end with this, so if you end up with code you can share that would be great!

components/workspace/src/polylith/clj/core/workspace/core.clj
```diff
@@ -31,14 +30,14 @@
         {:keys [top-namespace interface-ns color-mode]} settings
         suffixed-top-ns (common/suffix-ns-with-dot top-namespace)
         interfaces (interfaces/calculate components)
-        interface-names (apply sorted-set (filter identity (mapv :name interfaces)))
+        interface-names (into (sorted-set) (keep :name) interfaces)
```
tengstrand, line 33: Awesome that you find and improve things like this!

tengstrand (approved):

imrekoszo:
>  coverage

@SavageMessiah in case you use kaocha, I managed to get cloverage working through [polylith-kaocha](https://github.com/imrekoszo/polylith-kaocha/blob/f8536b683842ac1d70eb1e1ab918be0369c43ce5/projects/example/test-resources/polylith-kaocha/example/tests.edn#L3)

PR #210, opened 2022-04-10
feat: add name of tar-file to sha-file

TimoKramer:
This enables the use of sha256sum -c for easier check


TimoKramer:
Sorry, I just realized that there is the file-name missing. I should have seen this earlier. With the filename inside the sha-file you can use `sha256sum -c poly-0.2.14-alpha.tar.gz.sha256` for ease of use.

tengstrand (approved):

PR #220, opened 2022-05-09
Feature/issue 206

nkabir:
A pull request based on suggestions from @tengstrand 

There is still work to do. For example, I assume paths contain `src`, `test`, and `resources`. 

However, I'm still new to the codebase so I thought it may be helpful for others to see the changes and provide feedback.


tengstrand:
We decided to reject this PR.

Issue #217, opened 2022-04-28
Poly tool should check deps.edn for src and test path


nkabir:
**Describe the bug**
The poly tool currently assumes all source is under `src` and `test`. Changing `deps.edn` for a component does not affect this. This is unexpected behavior that leads to developer confusion.

**To Reproduce**
1. Create a component
2. Edit `deps.edn` to change `src` path to `src/clj`
3. See warning from `poly check` indicating that `clj.se` namespace is non-root

**Expected behavior**
Poly tool should use `deps.edn` and find source under `src/clj` rather than include `clj` in namespace.

nkabir:
This bug is related to "Use source paths from deps.edn files #206"

tengstrand:
Related to #206

tengstrand:
A fix has been merged to master. Please check it out @nkabir!

Issue #206, opened 2022-04-05
Use source paths from deps.edn files


tengstrand:
Don't read read source code from hard coded paths, see this [example](https://github.com/polyfy/polylith/blob/master/components/workspace-clj/src/polylith/clj/core/workspace_clj/projects_from_disk.clj#L152) (thanks @imrekoszo for pointing this out!).

Instead, use the paths in the `deps.edn` files.

Make sure we support more than one src or test directory (because it's supported by tools.deps).

imrekoszo:
Thanks for creating the issue! Let me try and add some extra thoughts.

The main area I would add more checks around is paths included in the development project and profiles vs _entities_ (bricks+projects) available in the workspace.

1. every entity that has sources or tests should be included in at least one profile or the development project
2. when an entity is included in the dev project or a profile, all of its source and test paths should be included (check top-level deps.edn vs entity deps.edn)
3. top-level deps.edn should not refer to any entity paths that are not present in the entity's own deps.edn

Also with regards to

> has a `test` directory, but it's not defined in its `deps.edn` file

this might trigger false positives (`doc` or `examples` etc. directories next to the entity deps.edn) so I'd keep the list of checked directory names short (src, test, resources) _and_ configurable in workspace.edn.

imrekoszo:
Here are the relevant locations in code where I found these hardcoded:

- https://github.com/polyfy/polylith/blob/abd20bb656e2cdd712732d80e1238316d3ea1b7f/components/workspace-clj/src/polylith/clj/core/workspace_clj/bases_from_disk.clj#L11-L12
- https://github.com/polyfy/polylith/blob/abd20bb656e2cdd712732d80e1238316d3ea1b7f/components/workspace-clj/src/polylith/clj/core/workspace_clj/components_from_disk.clj#L13-L14
- https://github.com/polyfy/polylith/blob/abd20bb656e2cdd712732d80e1238316d3ea1b7f/components/workspace-clj/src/polylith/clj/core/workspace_clj/projects_from_disk.clj#L151

The following are perhaps less relevant but still need checking:

- https://github.com/polyfy/polylith/blob/abd20bb656e2cdd712732d80e1238316d3ea1b7f/components/path-finder/src/polylith/clj/core/path_finder/interface/criterias.clj#L35-L42
[...]

PR #224, opened 2022-05-16
Use sources defined in deps.edn instead of hard coded values

tengstrand:
See issue #206 and #217


components/workspace-clj/src/polylith/clj/core/workspace_clj/bases_from_disk.clj
```diff
@@ -2,27 +2,32 @@
...
 
-(defn read-base [ws-dir ws-type user-home top-namespace ns-to-lib top-src-dir brick->non-top-namespaces base-name]
+(defn read-base [ws-dir ws-type user-home top-namespace ns-to-lib top-src-dir base-name]
   (let [base-dir (str ws-dir "/bases/" base-name)
-        src-dir (str base-dir "/src/" top-src-dir)
-        test-dir (str base-dir "/test/" top-src-dir)
-        namespaces (ns-from-disk/namespaces-from-disk src-dir test-dir)
         config (config-from-disk/read-config-file ws-type base-dir)
+        base-src-dirs (brick-dirs/top-source-dirs base-dir top-src-dir (-> config :paths))
+        base-test-dirs (brick-dirs/top-source-dirs base-dir top-src-dir (-> config :aliases :test :extra-paths))
```
imrekoszo:
```clojure
(-> config :paths))
(-> config :aliases :test :extra-paths))
```

I can see this pattern repeated also in components-from-disk. Any reason to not push it down to `brick-dirs` and only pass `config`?

components/workspace-clj/src/polylith/clj/core/workspace_clj/bases_from_disk.clj
```diff
@@ -2,27 +2,32 @@
...
+        namespaces (ns-from-disk/namespaces-from-disk base-src-dirs base-test-dirs)
         entity-root-path (str "bases/" base-name)
-        lib-deps (lib/brick-lib-deps ws-dir ws-type config top-namespace ns-to-lib namespaces entity-root-path user-home)]
+        lib-deps (lib/brick-lib-deps ws-dir ws-type config top-namespace ns-to-lib namespaces entity-root-path user-home)
+        source-paths (concat (-> config :paths)
+                             (-> config :aliases :test :extra-paths))
```
imrekoszo: Hmm the same pattern is repeated here and in components-from-disk as well. This might also be a policy ("path calculation considers top-level `:paths` and `:extra-paths` in the `:test` alias) that applies to all bricks (and to some extent maybe even projects). I would suggest extracting, although potentially not to `brick-dirs`.

components/workspace-clj/src/polylith/clj/core/workspace_clj/brick_dirs.clj
```diff
@@ -0,0 +1,10 @@
+(ns polylith.clj.core.workspace-clj.brick-dirs)
+
+(defn source-dirs [brick-dir paths]
+  (mapv #(str brick-dir "/" %)
+        (filter #(not= "resources" %)
+                paths)))
```
imrekoszo: To save a bit on needless allocation:

```suggestion
  (into []
        (comp (filter #(not= "resources" %))
              (map #(str brick-dir "/" %)))
        paths)
```

I do find it somewhat strange that this duplicates the definition of resource dirs from [components/path-finder/src/polylith/clj/core/path_finder/interface/criterias.clj](https://github.com/polyfy/polylith/pull/224/files?diff=unified&w=1#diff-83f5831a345e60a6d3bd0e5cf59db8dc93ef646e0a1acc59d85e6a1b55e26ca8).

It also doesn't seem like a perfect duplication, as the one in path-finder also considers "clj/resources" a resource dir.

I'm not familiar with this part really so wouldn't be able to tell whether it's correct this way or not.

components/workspace-clj/src/polylith/clj/core/workspace_clj/interface_defs_from_disk.clj
```diff
@@ -36,7 +36,8 @@
 
 (defn defs-from-disk
   "Example of a src-dir: ./components/workspace-clj/src/polylith/clj/core/workspace_clj"
-  [src-dir interface-ns]
+  [src-dirs interface-ns]
   (vec (sort-by (juxt :sub-ns :type :name params)
                 (mapcat #(interface-from-disk % interface-ns)
-                        (interface-namespaces src-dir interface-ns)))))
+                        (mapcat #(interface-namespaces % interface-ns)
+                                src-dirs)))))
```
imrekoszo: suggestion:

```clojure
(defn defs-from-disk
  "Example of a src-dir: ./components/workspace-clj/src/polylith/clj/core/workspace_clj"
  [src-dirs interface-ns]
  (->> src-dirs
       (into []
             (comp
              (mapcat #(interface-namespaces % interface-ns))
              (mapcat #(interface-from-disk % interface-ns))))
       (sort-by (juxt :sub-ns :type :name params))
       (vec)))
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/namespaces_from_disk.clj
```diff
@@ -70,31 +70,38 @@
 (defn imports [ns-statements]
   (vec (sort (mapcat import (filterv import? ns-statements)))))
 
+(defn skip-slash [path]
+  (if-let [skipped (str-util/skip-until path "/")]
+    skipped
+    path))
```
imrekoszo:
```suggestion
  (or (str-util/skip-until path "/") path))
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/namespaces_from_disk.clj
```diff
@@ -70,31 +70,38 @@
...
      :imports (-> content first imports)}))
 
-(defn source-namespaces-from-disk [root-dir]
-  (mapv #(->namespace root-dir %)
-        (-> root-dir
+(defn source-namespaces-from-disk [source-dir]
+  (mapv #(->namespace source-dir %)
+        (-> source-dir
             file/paths-recursively
             common/filter-clojure-paths)))
```
imrekoszo, line 97: Just taking advantage of threading:

```suggestion
(defn source-namespaces-from-disk [source-dir]
  (->> source-dir
       file/paths-recursively
       common/filter-clojure-paths
       (mapv #(->namespace source-dir %))))
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/namespaces_from_disk.clj
```diff
@@ -70,31 +70,38 @@
...
 
-(defn namespaces-from-disk [src-dir test-dir]
-  (let [src (source-namespaces-from-disk src-dir)
-        test (source-namespaces-from-disk test-dir)]
+(defn namespaces-from-disk [src-dirs test-dirs]
+  (let [src (vec (mapcat #(source-namespaces-from-disk %)
+                         src-dirs))
+        test (vec (mapcat #(source-namespaces-from-disk %)
+                          test-dirs))]
```
imrekoszo, line 103: Reducing allocations:
```suggestion
  (let [src (into [] (mapcat source-namespaces-from-disk) src-dirs)
        test (into [] (mapcat source-namespaces-from-disk) test-dirs)]
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/non_top_namespace.clj
```diff
@@ -4,35 +4,35 @@
...
+(defn non-top-namespaces-for-source [brick-type brick-name brick-dir top-src-dir source-dir]
+  (let [path (str brick-dir "/" source-dir)
+        path-with-slash (str path "/")
+        paths (mapv #(str-util/skip-prefix (str %) path-with-slash)
+                    (filterv #(and (-> % file/directory? not)
+                                   (non-data-reader-file? (str %)))
+                             (file/files-recursively path)))]
+    (mapv #(non-top-ns-map brick-type brick-name source-dir %)
+          (filter #(is-not-top-ns? % top-src-dir)
+                  paths))))
```
imrekoszo: Reducing allocations:

```suggestion
(defn non-top-namespaces-for-source [brick-type brick-name brick-dir top-src-dir source-dir]
  (let [path (str brick-dir "/" source-dir)
        path-with-slash (str path "/")]
    (->> path
         (file/files-recursively)
         (into []
               (comp
                (filter #(and (not (file/directory? %))
                              (non-data-reader-file? (str %))))
                (map #(str-util/skip-prefix (str %) path-with-slash))
                (filter #(is-not-top-ns? % top-src-dir))
                (map #(non-top-ns-map brick-type brick-name source-dir %)))))))
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/non_top_namespace.clj
```diff
@@ -4,35 +4,35 @@
...
-                                      (common/filter-clojure-paths
-                                        (map #(str-util/skip-prefix % (str ws-dir "/"))
-                                             (file/paths-recursively ws-dir))))))]
-    (reduce add-non-top-ns {} non-nss)))
+(defn non-top-namespaces [brick-type brick-name brick-dir top-src-dir source-paths]
+  (let [namespaces (mapcat #(non-top-namespaces-for-source brick-type brick-name brick-dir top-src-dir %)
+                           (filter #(not= "resources" %)
+                                   source-paths))]
+    (when (seq namespaces)
+      (vec namespaces))))
```
imrekoszo, line 40: I noticed that the magic string `"resources"` appears again and would suggest considering instead referring to a policy here.

Some more reduction of allocations:

```suggestion
(defn non-top-namespaces [brick-type brick-name brick-dir top-src-dir source-paths]
  (let [namespaces (into []
                         (comp
                          (filter #(not= "resources" %))
                          (mapcat #(non-top-namespaces-for-source
                                    brick-type brick-name brick-dir top-src-dir %)))
                         source-paths)]
    (when (seq namespaces)
      namespaces)))
```

components/workspace-clj/src/polylith/clj/core/workspace_clj/project_paths.clj
```diff
@@ -0,0 +1,20 @@
+(ns polylith.clj.core.workspace-clj.project-paths
+  (:require [polylith.clj.core.path-finder.interface.criterias :as c]
+            [polylith.clj.core.path-finder.interface.extract :as extract]
+            [polylith.clj.core.path-finder.interface.select :as select]))
+
+(defn source-paths [project-name is-dev src-paths]
+  (if is-dev src-paths
+             (mapv #(str "projects/" project-name "/" %)
+                   src-paths)))
```
imrekoszo:
```suggestion
(defn source-paths [project-name is-dev src-paths]
  (cond->> src-paths
           (not is-dev) (mapv #(str "projects/" project-name "/" %))))
```

components/path-finder/src/polylith/clj/core/path_finder/interface/criterias.clj
```diff
@@ -32,18 +32,20 @@
 (defn not-exists? [{:keys [exists?]}]
   (not exists?))
 
-(defn src-path? [{:keys [path]}]
-  (str/ends-with? path "/src"))
+(defn resources? [path]
+  (or (= path "resources")
+      (str/ends-with? path "/resources")))
```
imrekoszo, line 37: I am curious: why are resource paths handled differently in Polylith? Clojure doesn't seem to treat resource and source paths differently as I learned through completely unrelated issues.

For example, Clojure and related tools originating from the core team consider `.clj` files that are under `"resources"` candidates for loading.

(See [this issue](https://github.com/clj-kondo/clj-kondo/issues/1685) and Alex Miller's comments [here](https://ask.clojure.org/index.php/11434/could-namespace-refresh-sophisticated-directory-filtering?show=11437#c11437) and [here](https://ask.clojure.org/index.php/11434/could-namespace-refresh-sophisticated-directory-filtering?show=11882#c11882))

components/path-finder/src/polylith/clj/core/path_finder/sources_on_disk.clj
```diff
@@ -3,8 +3,8 @@
 
 (defn entity-source-paths [ws-dir entity-type entity]
   (map #(str entity-type "/" entity "/" %)
-    (filter #(contains? #{"src" "resources" "test"} %)
-      (file/directories (str ws-dir "/" entity-type "/" entity)))))
+    (filter file/not-hidden?
```
imrekoszo: Noticed the nested lazy seqs here and looked at the file. I would refactor the file to reduce allocations as:

```clojure
(ns polylith.clj.core.path-finder.sources-on-disk
  (:require [polylith.clj.core.file.interface :as file]))

(defn- entity-source-paths [ws-dir entity-type entity]
  (eduction
   (filter file/not-hidden?)
   (map #(str entity-type "/" entity "/" %))
   (file/directories (str ws-dir "/" entity-type "/" entity))))

(defn- entity-paths [ws-dir entity-type]
  (eduction
   (mapcat #(entity-source-paths ws-dir entity-type %))
   (file/directories (str ws-dir "/" entity-type))))

(defn source-paths [ws-dir]
  (->> ["bases" "components" "projects"]
       (into [] (mapcat #(entity-paths ws-dir %)))
       (sort)
       (vec)))
```

imrekoszo (commented):
Nice changes! I had a few questions and suggestions as I read the PR, but as I'm not fully in context, feel free to ignore any or all of them.

components/path-finder/src/polylith/clj/core/path_finder/interface/criterias.clj
tengstrand: This is only based on the idea that people would prefer to put their source code in the `src` directory (and similar) and leave `resources` directory to other type of files, like images and such. If people follow this pattern, then I believe it has a value, because it brings some order in my mind (this is how you would separate it in Java). But if people want to put source code in their `resources` directory, then things would go wrong the way the poly tool handles this, so I understand why you ask. I don't know if this is a problem though, because no one has complained about it so far.

imrekoszo: I also tend to follow that convention and agree with you on the separation by file type. Based on what I gathered from Alex however is that tools shouldn't really make a distinction between source and resource types.

This is why I'm asking the significance of this. I haven't found this mentioned in the poly docs so I was wondering if you could point me to where it matters for poly whether something is under src or resources?

Issue #227, opened 2022-05-26
Support XDG systems properly for config.edn file


seancorfield:
**Is your feature request related to a problem? Please describe.**

On XDG systems, the convention is to put user-level config in directories under `$XDG_CONFIG_HOME/` (typically `$HOME/config/`) so Polylith's `config.edn` should be in `$XDG_CONFIG_HOME/polylith/config.edn` instead of `$HOME/.polylith/config.edn`.

**Describe the solution you'd like**

Polylith should detect the `XDG_CONFIG_HOME` environment variable and use the appropriate path on such systems, but continue to use the other location when `XDG_CONFIG_HOME` is not set.

To avoid backward compatibility issues, Polylith should auto-migrate the current `config.edn` to the correct location on XDG systems and replace the old `config.edn` contents with a comment indicating the new location.

Specifically:
* on a non-XDG system, continue to use `$HOME/.polylith/config.edn` as currently (creating it if missing)
* on an XDG system:
  * if `$XDG_CONFIG_HOME/polylith/config.edn` does not exist,
    * if `$HOME/.polylith/config.edn` exists
      * copy it to the XDG location, then replace the old content with a comment indicating the XDG location
    * if `$HOME/.polylith/config.edn` does not exist
      * create the XDG config with the default content
      * perhaps also create `$HOME/.polylith/config.edn` with a comment indicating the XDG location? @imrekoszo cautioned that XDG systems are about not cluttering up the `$HOME` directory!
  * use `$XDG_CONFIG_HOME/polylith/config.edn`

**Describe alternatives you've considered**

1. Leave the behavior as it is -- XDG users may continue to complain that Polylith does not follow conventions.
2. Use/create `config.edn` in the XDG location and ignore any old file under `$HOME/.polylith` -- XDG users might still edit the non-XDG file (e.g., if they're following an outdated blog post) and wonder why their changes don't work; this could also be a breaking change for users that have already configured Polylith using the "old" `config.edn` since their changes would stop working when they upgraded `poly` and it created a new, default XDG `config.edn` file.

**Additional context**

Polylith's own CI assumes the `$HOME/.polylith/config.edn` location. If `poly` auto-migrates on XDG systems, this setup does not need to change even if the CI system is XDG.

Polylith calculates the path to `config.edn` in many locations in the code. This should all be centralized into a single `component` (and some of the existing functionality in `components/command` migrated to this new component).

seancorfield:
[...]

PR #228, opened 2022-05-28
Address issue 227 by adopting XDG config convention

seancorfield:
This addresses #227 by:

* Changing the default `config.edn` location to the XDG location (default `~/.config/polylith/config.edn` but overridable via the `XDG_CONFIG_HOME` environment variable).
* Adding a migration of `~/.polylith/config.edn` to the new location, adding a comment in the old file so any projects still using older `poly` versions can still read it.

> Note: I'm not sure how Circle CI will handle environment variable syntax in `config.yml` but I've tried to make it work regardless of `XDG_CONFIG_HOME` being set; If that breaks it, just changing to `~/.config/polylith` in those four lines should solve it.


seancorfield:
Apparently, the Circle CI updates were acceptable. ~The conflicts are cosmetic and seem to reflect a difference due to the pluggable test runner change (which I thought was on master and I _thought_ I was all merged up to date but apparently not -- LMK if you want me to fix the PR).~ PR updated after full rebasing.

tengstrand (changes requested):
Hi Sean!
This looks great to me. I have just tested it on my machine, which is not an XGD system, and it seems to work as before. I will approve this if you update polylith.clj.core.version.interface with the correct issue and date, e.g.:

(def revision "alpha-issue227-01")
(def date "2022-05-29")

I need to update the gitbook documentation also, branch "Release 0.2.15" (that's the downside of letting the documentation live in a separate repo).

seancorfield:
Updates made.

tengstrand (approved):
Thanks!

PR #232, opened 2022-06-06
return non-zero exit code if test setup fails

### LINKED ISSUE #229: Poly test should return non-zero exit code when per-project test-setup has failed
**Describe the bug**
`poly test` should return non-zero exit code when per-project `:setup-fn` is given and failed (throws an exception).

**To Reproduce**
1. Have a per-project `:setup-fn` function set in `workspace.edn` and make it throw an exception.
2. Run `poly test project:{the project name}; echo "exit code: $?"` 
3. You can see a message in red `Test setup failed: *****` and `exit code: 0`

**Expected behavior**
Poly test should return non-zero exit code.

**Operating System (please complete the following information):**
 - OS: MacOS
 - Version: 12.0.1

**Versions (please complete the following information):**
 - Java: java 17.0.2 
 - Poly: 0.2.14-alpha
 - Brew: 3.3.9-80-g26694fd

devgrapher:
Fix #229 

This PR is just to give you a hint to fix the issue I reported.
It seems to fix the issue, but I'm not sure if this is the right way to fix it since I'm not used to the codebase.

### DIFF
--- a/components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
+++ b/components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
@@ -97,7 +97,7 @@
             all-paths (into #{} cat [(:src paths) (:test paths) lib-paths])
             class-loader (common/create-class-loader all-paths color-mode)
             setup!* (delay (execute-fn setup-fn "setup" name class-loader color-mode))
-            setup-failed? #(and (realized? setup!*) (not (deref setup!*)))
+            setup-failed? #(not (deref setup!*))
             setup-succeeded? #(and (realized? setup!*) (deref setup!*))
             teardown!* (delay (execute-fn teardown-fn "teardown" name class-loader color-mode))
             runner-opts (merge opts
--- a/components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
+++ b/components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
@@ -64,11 +64,12 @@
              (test-runner-contract/test-runner-name test-runner))]
     (if-not (test-runner-contract/tests-present? test-runner runner-opts)
       (println (str "No tests to run " for-project-using-runner "."))
-      (when (deref setup-delay)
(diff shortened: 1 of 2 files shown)


components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
```diff
@@ -97,15 +97,16 @@
             all-paths (into #{} cat [(:src paths) (:test paths) lib-paths])
             class-loader (common/create-class-loader all-paths color-mode)
             setup!* (delay (execute-fn setup-fn "setup" name class-loader color-mode))
-            setup-failed? #(and (realized? setup!*) (not (deref setup!*)))
+            setup-failed? #(not (deref setup!*))
```
imrekoszo: With this change, setup will run in line 108 even when there are no test runners that find any tests (see the implementation of `run-tests-for-project-with-test-runner`). I would probably suggest making line 67 an `if` and throwing on its else branch. What do you think? Is there something I'm missing?

tengstrand (changes requested):
Also update :
https://github.com/polyfy/polylith/blob/7c6c94ada0a680476e74b64e903cd4b57783bd26/components/version/src/polylith/clj/core/version/interface.clj#L7
To `alpha-issue229-01` and the date:
https://github.com/polyfy/polylith/blob/7c6c94ada0a680476e74b64e903cd4b57783bd26/components/version/src/polylith/clj/core/version/interface.clj#L10

This is a "service" for people who execute the `poly` tool via a hash, so that the version can be checked by executing `poly version`.

components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
devgrapher: Yes, you're totally right. I'll move the `if`. Thanks!

imrekoszo (dismissed):

tengstrand:
Looks like we can merge this now @imrekoszo?

imrekoszo:
> Looks like we can merge this now

I'm okay with that!

Issue #234, opened 2022-06-15
Test setup failure no longer aborts test run


seancorfield:
**Describe the bug**
It looks like the integration of pluggable test runners has broken some of the assumptions around project test fixtures:

```
Running test setup for the publisher project: worldsingles.application.fixtures/pre-test

Test setup failed: Syntax error macroexpanding at (worldsingles/user.clj:1:1).
Running test setup for the system project: ws.system.fixtures/pre-test
...
```

If the "Test setup failed", the entire test run should abort. It used to abort, and it no longer does.

**To Reproduce**
Add a project-specific test setup function that throws an exception (for a project that will be among the first to be tested).

Run `poly test` in a way that runs tests for multiple projects, including that one.

**Expected behavior**
The test setup for that project will fail (b/c it throws an exception) and it _should abort the test run_.

seancorfield:
@imrekoszo I haven't looked at the PR in detail that introduced this but I suspect it catches and reports an exception around test running, but no longer propagates it to the caller?

seancorfield:
Also: need to ensure that if test setup _succeeds_, then test teardown _must be run even if the tests fail_.

tengstrand:
Add tests, see [comment](https://clojurians.slack.com/archives/C013B7MQHJQ/p1655355687828689?thread_ts=1655333868.287389&cid=C013B7MQHJQ) in Slack: 
- If test setup fails, the whole test run should stop (and test teardown should not happen)
- If test setup succeeds and the tests pass, the test teardown should run and if it fails, the whole test run should stop
- If test setup succeeds but the tests themselves fails, the test teardown should still happen -- and if it succeeds, behave according to the test failures
- If test setup succeeds but the tests themselves fails, the test teardown should still happen -- but if that fails, the whole test run should stop

imrekoszo:
Thank you @seancorfield and apologies for the breakage. Could you please share the version (sha) of poly you are using to make sure we're looking at the correct one?

Also, does reverting poly to `28cca05514500fe995d4107dd5649c162b57b02a` or earlier fix this?

Here's how it's done in `28cca005` (just before pluggable TRs):
https://github.com/polyfy/polylith/blob/28cca05514500fe995d4107dd5649c162b57b02a/components/test-runner/src/polylith/clj/core/test_runner/core.clj#L121-L125

And after pluggable TRs:
[...]

PR #243, opened 2022-08-06
Issue 234

tengstrand:
Step 50 to 55 in the script [create-example.sh](https://github.com/polyfy/polylith/blob/issue-234/scripts/create-example.sh) has been updated and executes different `poly` commands against the projects in the workspace [for-test](https://github.com/polyfy/polylith/tree/issue-234/examples/for-test) to verify that issue #234 works correctly.


components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
```diff
@@ -163,17 +181,12 @@
...
           (print-bricks-to-test component-names base-names bricks-to-test color-mode)
           (println)
-          (doseq [{:keys [name] :as project} projects-to-test]
-            (->> {:project project
-                  :test-settings (get-in settings [:projects name :test])}
-                 (merge run-opts)
-                 (run-tests-for-project)))))
+          (reduce run-tests-for-project true
+                  (map #(test-opts workspace settings changes % is-verbose color-mode)
+                       projects-to-test))))
```
imrekoszo: There should be no need for the extra arity in `run-tests-for-project`. I think the don't-run-if-failures-have-been-seen logic belongs here and not there. One way to achieve it would be:

```suggestion
          (transduce
           (comp (map #(test-opts workspace settings changes % is-verbose color-mode))
                 (map run-tests-for-project))
           (completing (fn [_ x] (cond-> x (not x) (reduced))))
           true
           projects-to-test)))
```

tengstrand: I agree, so I updated to your suggestion @imrekoszo!

PR #246, opened 2022-08-11
fix #245 by initializing Clojure runtime properly

### LINKED ISSUE #245: Isolated classloader context for poly test is unusual/incorrect
See https://github.com/seancorfield/poly-classloader-bug for repro case and instructions.

This is blocking our migration at work. Although we can add an explicit require to get around this initial failure, other tests subsequently fail for classloader-related issues so it seems like solving this initial problem might address some larger issues.

Although the generated project for the repro uses an older version of Polylith, this behavior continues to exist in the latest release.

seancorfield:
Loading `clojure.lang.RT` doesn't fully initialize the Clojure runtime. This PR switches Polylith's classloader to the "new" (as of Clojure 1.6!) Java API to require `clojure.main` and get the classloader prepared.

### DIFF
--- a/components/common/src/polylith/clj/core/common/class_loader.clj
+++ b/components/common/src/polylith/clj/core/common/class_loader.clj
@@ -5,7 +5,7 @@
   (:import (java.net URLClassLoader URL)))
 
 (def base-classloader
-  (or (.getClassLoader clojure.lang.RT)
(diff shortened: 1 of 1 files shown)


furkan3ayraktar (approved):
Thanks for this! 🎉

tengstrand (approved):

Issue #260, opened 2022-11-14
External process test runner


seancorfield:
**Is your feature request related to a problem? Please describe.**
Polylith's classloader isolation approach is problematic for a number of reasons (issues and PRs have already been submitted/merged to mitigate some of these problems, including excessive heap usage on very large projects). Running each project's test suite in an isolated _subprocess_ would address all of these issues, once and for all.

**Describe the solution you'd like**
A new test runner, that `poly test` can invoke as a subprocess, to run a project's test suite in complete isolation. This requires some minor modifications to Polylith to support this:
* Instead of running the setup/teardown for a project once and then running multiple test runners on the project, run the setup/teardown for each test runner -- this is a prerequisite for an external process since the setup/teardown must run in the same context as the tests, i.e., in the subprocess.
* Modify the (currently unconditional) construction of the classloader such that it can be deferred until a test runner seems to need it -- also a prerequisite for an external process since it wouldn't need the classloader (only the classpath) and repeatedly building the classloader is part of what causes the excessive heap issues in very large projects.
* Provide a way for a test runner to indicate that it wishes to be run as a subprocess via a classpath: it would also be responsible for running the setup and teardown functions.
* Design a _simple_ command-line API for such test runners -- proposal: pass the color-mode string, the project-name, the setup-fn name if present, the list of namespaces to require and run tests from, and the teardown-fn name if present.

**Describe alternatives you've considered**
Attempts to mitigate classloader isolation and memory usage have already been applied and are insufficient.

**Additional context**
We are using my fork of Polylith with an external test runner at work and it addresses all our issues and seems to otherwise function identically to the current in-process test runner. A PR for a cleaned-up version of this will be forthcoming.

PR #262, opened 2022-11-15
Adds support for an external test runner per #260

seancorfield:
Minimal changes to support #260 

This does not include the external test runner itself, which can be a separate project. This does not provide any documentation on the new `ExternalTestRunner` protocol or how to build other external test runners. I don't know what was added to the Polylith docs or tool repo when the pluggable test runner machinery was added, but I expect that something similar will need adding by way of documentation for this?

See https://github.com/seancorfield/polylith-external-test-runner


seancorfield:
I'm closing this in favor of #263 from @furkan3ayraktar

PR #263, opened 2022-11-20
External Test Runner Support

### LINKED ISSUE #252: Memory consumption is very high on large projects
**Describe the bug**
This is a bit vague -- sorry -- but our project at work has gotten large enough (19 projects, 107 components) that when a low-level component (or `deps.edn` file) is updated and it triggers a test run of almost all projects and all components, our CI job gets killed for memory usage (BitBucket, 2x instance, already the largest we can have).

**To Reproduce**
Very hard to reproduce!

**Expected behavior**
`poly test` should free up heap and non-heap that it doesn't use (tricky: I suspect at least some of this is the large classloader space created by so many projects and so many components, but haven't proved that yet).

**Partial Workaround**
We're running a fork of Polylith that has a call to `(System/gc)` in the test orchestrator after each test suite has run. This only signals to the JVM that we'd like garbage collection as a hint, but it does seem to have been sufficient to reduce the memory footprint _most of the time_. See https://github.com/seancorfield/polylith/compare/4e946eab081cdef317377da2e9b2f61158dbc422..as-alias 

We may continue to tweak min/max heap sizes via JVM options to see if we can get a better outcome, but we've already had to set min/max to control heap usage on other environments so we're threading a needle here already. I don't know if there's an "ideal" solution to this (but Polylith's classloader isolation is problematic in a couple of areas as noted in other discussions on Slack and other GH issues).

### LINKED ISSUE #253: Classloader isolation has "leaks"
**Describe the bug**
Code that uses global JVM thread pools (such as code based on fork/join) can end up loading classes into those threads which then conflict with those same classes loaded into the classloaders `poly test` creates to run code.

**To Reproduce**
Exceedingly hard to reproduce(!) and even in some repro cases, it is intermittent. We've managed to mitigate _most_ instances by moving `reify` calls into the main thread where possible in our code (so no new class instances get loaded in the fork/join thread classloader) but there are places where we just can't do this.

**Expected behavior**
Given the use of isolated classloaders, this _is_ the expected behavior -- it's just problematic. Running tests in a subprocess would resolve this (and other classloader issues) but would make Polylith's test runner logic much more complex I suspect.

**Partial Workaround**
In addition to reorganizing our code as described above, we have found that changing the `*use-context-classloader*` binding to `false` seems to help resolve more cases of this issue. The default is `true` so I'm not sure why that `binding` was ever introduced, but at least it provides for a smaller "fix": https://github.com/seancorfield/polylith/commit/5945bd2cf5b36f07a759744f1de702c79970a9b4

Since we made that change on my fork, we have not run into the ClassCastException that the cross-classloader issue described above was causing -- so I think it's a safe change to make and will help others who might be using fork/join code.

### LINKED ISSUE #260: External process test runner

furkan3ayraktar:
Another stab at the issue #260 

This PR aimed to develop a cleaner solution with fewer changes to the existing code. It is a result of my discussion with @tengstrand and is heavily inspired by @seancorfield's PR #262. 

See [seancorfield/polylith-external-test-runner](https://github.com/seancorfield/polylith-external-test-runner) as an example of an external test runner.

There is no documentation about custom test runners, except the doc string in the `test-runner-contract` interface. It would be nice to add a new page to the poly documentation and explain how to develop custom in-process and external test runners.

Resolves #260, resolves #252, resolves #253
Also closes #262

### DIFF
--- a/components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
+++ b/components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
@@ -47,10 +47,20 @@
   ```
   (defn create [{:keys [workspace project test-settings is-verbose color-mode changes]}]
     ...
-    (reify TestRunner ...)
 
-    ; Optional, only if you want an external test runner
-    (reify ExternalTestRunner ...))
+    (reify
+      test-runner-contract/TestRunner
+      (test-runner-name [this] ...)
+
+      (test-sources-present? [this] ...)
+
+      (tests-present? [this runner-opts] ...)
+
+      (run-tests [this runner-opts] ...)
+
+      ; Optional, only if you want an external test runner
+      test-runner-contract/ExternalTestRunner
+      (external-process-namespace [this] ...)))
   ```
 
   `workspace` passed to the constructor will contain `:user-input`, which
@@ -58,12 +68,12 @@
 
   Add your constructor function in the workspace.edn:
 
-  {:test {:create-test-runner my.namespace/create} ; to use it globally
+  {:test {:create-test-runner [my.namespace/create]} ; to use it globally
 
    :projects {; to use it only for a project
-              \"project-a\" {:test {:create-test-runner my.namespace/create}}
+              \"project-a\" {:test {:create-test-runner [my.namespace/create]}}
               ; to reset the global setting to default
-              \"project-b\" {:test {:create-test-runner :default}}}}"
+              \"project-b\" {:test {:create-test-runner [:default]}}}}"
 
   (test-runner-name [this]
     "Returns a printable name that the poly tool can print out for
--- a/components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
+++ b/components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
@@ -3,66 +3,103 @@
 (defprotocol TestRunner
   "Implement this protocol to supply a custom test runner.
 
-  `test-runner-name`
-    - should return a printable name that the test orchestrator can print out for information purposes
+  Runner options:
+  `is-verbose`      -> A boolean indicating if we are running in verbose mode
+                       or not. TestRunner can use this to print additional
+                       information about the test run.
+  `color-mode`      -> The color-mode that the poly tool is currently running with.
+                       TestRunner is expected to respect the color mode.
+  `project`         -> A map containing the project information.
+  `all-paths`       -> A vector of all paths necessary to create a classpath for
+                       running the tests in isolation within the context of the
+                       current project.
+  `setup-fn`        -> An optional setup function for tests defined in the
+                       workspace config. The poly tool will run this function
+                       before calling run-tests only if this is an in-process
+                       TestRunner. If this is an ExternalTestRunner, the external
+                       test runner should run the setup-fn.
+  `teardown-fn`     -> An optional teardown function for tests defined in the
+                       workspace config. The poly tool will run this function
+                       after the run-tests function completes (exception or not),
+                       only if this is an in-process TestRunner. If this is an
+                       ExternalTestRunner, the external test runner should run
+                       the teardown-fn.
 
-  `test-sources-present?`
-    - called first
-    - if falsey, we short-circuit, not even the project classloader will be created
+  Additional options for in-process TestRunner:
+  `class-loader`    -> The isolated classloader created from the `all-paths`.
+                       This classloader will be used to evaluate statements within
+                       the project's context. Use this if you need more granular
+                       access. `eval-in-project` should be sufficient for most
+                       cases.
+  `eval-in-project` -> A function that takes a single form as its argument and
+                       evaluates it within the project's classloader. It returns
+                       the result of the evaluation. This is the primary interface
+                       for running tests in the project's isolated context.
 
-  `tests-present?`
-    - if falsey, run-tests won't be called; can eval forms in the project context
+  Additional options for ExternalTestRunner:
+  `process-ns`      -> The main namespace of the external test runner. This
+                       namespace will be invoked as a Java subprocess.
 
-  `run-tests`
-    - should throw if the test run is considered failed
+  Usage:
+  Create a constructor function that returns an instance of TestRunner or
+  ExternalTestRunner:
 
-  Special args:
+  ```
+  (defn create [{:keys [workspace project test-settings is-verbose color-mode changes]}]
+    ...
+    (reify TestRunner ...)
 
-  `eval-in-project`
-    - a function that takes a single form which it evaluates in the project classloader and returns its result
-    - this is the primary interface for running tests in the project's context
+    ; Optional, only if you want an external test runner
+    (reify ExternalTestRunner ...))
+  ```
 
-  `class-loader`
-    - the project classloader in case more granular access is needed to it
+  `workspace` passed to the constructor will contain `:user-input`, which
+  can be used to receive additional parameters for runtime configuration.
 
+  Add your constructor function in the workspace.edn:
 
-   To use a custom test runner, create a constructor that returns an instance of it:
+  {:test {:create-test-runner my.namespace/create} ; to use it globally
 
-   (defn create [{:keys [workspace project changes test-settings]}]
-     ,,,
-     (reify TestRunner ,,,))
+   :projects {; to use it only for a project
+              \"project-a\" {:test {:create-test-runner my.namespace/create}}
+              ; to reset the global setting to default
+              \"project-b\" {:test {:create-test-runner :default}}}}"
 
-   `workspace` passed to the constructor will contain `:user-input`, which
-   can be used to receive additional parameters for runtime configuration.
+  (test-runner-name [this]
+    "Returns a printable name that the poly tool can print out for
+    information purposes")
 
-   And in workspace.edn:
+  (test-sources-present? [this]
+    "The poly tool calls this first before attempting to run any tests. If
+    it returns a falsy value, we short-circuit. Not even the project
+    classloader will be created")
 
-   {:test {:create-test-runner my.namespace/create} ;; to use it globally
+  (tests-present? [this runner-opts]
+    "The poly tool calls this before calling the run-tests. If it returns a
+    falsy value, run-tests won't be called. The runner-opts passed to this
+    function is identical to the one passed to the run-tests. It can evaluate
+    forms in the project's context.")
 
-    :projects {\"project-a\" {:test {:create-test-runner my.namespace/create}} ;; to use it only for a project
-               \"project-b\" {:test {:create-test-runner :default}} ;; to reset the global setting to default
-               }}"
-  (test-runner-name [this])
-  (test-sources-present? [this])
-  (tests-present? [this {:keys [class-loader eval-in-project] :as opts}])
-  (run-tests [this {:keys [class-loader color-mode eval-in-project is-verbose] :as opts}]))
+  (run-tests [this runner-opts]
+    "It should run the tests and throw an exception if the test run is considered
+     failed."))
 
 (defprotocol ExternalTestRunner
-  "Extends the `TestRunner` protocol to provide an external process namespace for a test runner. Polylith uses
-  a classloader approach to run tests in isolation by default. `ExternalTestRunner` skips the classloaders and
-  uses `java` subprocesses.
+  "Extends the `TestRunner` protocol to provide an external process namespace
+  for a test runner. Polylith uses a classloader approach to run tests in
+  isolation by default. `ExternalTestRunner` skips the classloaders and uses
+  Java subprocesses."
 
-  `external-process-namespace`
-    - if returns nil (default), tests are run in isolated classloaders within the same process,
-    - if returns non-nil, it must be a symbol or string identifying the main namespace of an
-      external test-runner to be used.
-      - when an external test-runner is used, no classloader will be created
-        and the `setup-fn`/`teardown-fn` should be run by that test-runner
-        instead of by `poly` itself; the main namespace will be invoked as a
-        `java` subprocess with arguments determined by the test runner;
-      - the external test runner's `run-tests` function is passed:
-        - :process-ns -- the name of the main namespace as above
-        - :setup-fn -- the fully-qualified name of the project setup function
-        - :teardown-fn -- similarly for the teardown function
-        - :all-paths -- a sequence of all the elements of the classpath"
-  (external-process-namespace [this]))
+  (external-process-namespace [this]
+    "Returns a symbol or string identifying the main namespace of an external
+    test runner. If it returns nil (default), the test runner will be an
+    in-process test runner and the tests will run in an isolated classloader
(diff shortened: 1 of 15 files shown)


seancorfield:
That does look cleaner and more "idiomatic" in keeping with how the existing test runner is constructed.

Have you tested it works with my external test runner as-is? (curious as to whether it would cause any changes in my project; I can test this out tomorrow at work)

furkan3ayraktar:
I've run the Polylith project's test using your external test runner, and it worked as-is. I did not intentionally change any input that goes to the test runner, so I assume it should work with your project at work. Let me know if I caused any issues.

furkan3ayraktar:
I've found two issues with the test runner and addressed them with two commits.

**Issue 1**
If a project does not exist in the `workspace.edn` file, the tests did not run for that project even though its name was logged during the test run. The reason is that the project missing from the `workspace.edn` file did not get a default test runner in its settings; hence, no tests ran.

[This commit](https://github.com/polyfy/polylith/pull/263/commits/0628224b4e29671f39fc41463b377675e2a5b63b) fixes the issue.

**Issue 2**
The `run` function in the `test-runner-orchestrator` component stopped executing tests in other projects if the `run-tests-for-project` returned a falsy value for a project. The `run-tests-for-project` returns `nil` if there are no test runners in the project settings that see test sources. Hence it returns `nil` if there are no test runners for the project. I noticed this connected to issue 1. Even though fixing issue 1 ensures there will always be a test runner for a project, it is still the correct behavior to execute tests for projects with test runners seeing test sources even though one of the projects doesn't run any tests.

[This commit](https://github.com/polyfy/polylith/pull/263/commits/f0a93b529a60f1376c049becb3a17dce0c6a0be6) fixes the issue.

components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
```diff
@@ -56,19 +56,48 @@
                           " project: " e))
             (throw e))))))
 
+(defn execute-setup-fn [project color-mode {:keys [setup-fn runner-ns class-loader]}]
```
seancorfield: You have `runner-ns` in both `execute-*-fn` functions, but the key is `:process-ns` in `run-tests-for-project`, so this attempts to run the setup function without a class loader and blows up.

seancorfield:
Per the comment above, this breaks at work because we use `:setup-fn`/`:teardown-fn` in all of our projects and I'm guessing their are no tests around that for the external test runner. When that's fixed, I'll test the PR again at work.

components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
furkan3ayraktar: Ah, right! I shouldn’t commit at 1 am. I’ll fix it and add tests for those functions. I’ll let you know. Thanks for testing it out.

furkan3ayraktar: I've fixed this here: https://github.com/polyfy/polylith/pull/263/commits/010d5e853247cdaf0d9e6c6079b5aca64a86db98

furkan3ayraktar:
I've fixed the issue in the comment above. Let me know if you find further problems.

seancorfield:
That seems to work now! Thank you! I'll switch us over to this at work "full-time" until it is merged and let you know if any other issues come up.

components/workspace/src/polylith/clj/core/workspace/settings.clj
```diff
@@ -8,12 +8,24 @@
 (defn undefined-project [index project-name]
   [project-name {:alias (str "?" (inc index))}])
 
+(def default-test-runner
```
imrekoszo: There is similar code in `polylith.clj.core.workspace-clj.project-settings` for projects that are in workspace.edn. With this code here I'm wondering whether it would make sense to move all of that old code into this namespace so they are in one place?

imrekoszo: Nice catch by the way, looks like I should have implemented setting the default in this file as opposed to the other namespace.

components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
```diff
@@ -47,23 +47,33 @@
...
 
-  {:test {:create-test-runner my.namespace/create} ; to use it globally
+  {:test {:create-test-runner [my.namespace/create]} ; to use it globally
 
    :projects {; to use it only for a project
-              \"project-a\" {:test {:create-test-runner my.namespace/create}}
+              \"project-a\" {:test {:create-test-runner [my.namespace/create]}}
               ; to reset the global setting to default
-              \"project-b\" {:test {:create-test-runner :default}}}}"
+              \"project-b\" {:test {:create-test-runner [:default]}}}}"
```
imrekoszo: suggestion: add another example project here showing that multiple runners can also be provided in the vector

imrekoszo: The value of `:create-test-runner` doesn't have to be a vector by the way - `:default` and `[:default]` result in the same behavior, the former gets wrapped in a vector if that's how it's provided: https://github.com/polyfy/polylith/blob/abd20bb656e2cdd712732d80e1238316d3ea1b7f/components/workspace-clj/src/polylith/clj/core/workspace_clj/project_settings.clj#L18

Not sure if it's worth its own example here.

imrekoszo:
I checked this with polylith-kaocha and found no breakages.

components/workspace/src/polylith/clj/core/workspace/settings.clj
furkan3ayraktar: Yeah, I noticed this when I added a new project at work but forgot to add it to the `workspace.edn` file 😂 

I didn’t know where to put the shared code; it seemed like the two places did different things, and merging them would be wrong. What do you think @tengstrand?

components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
```diff
@@ -108,26 +137,26 @@
...
-            runner-opts (merge opts
-                               {:class-loader class-loader
-                                :eval-in-project (->eval-in-project class-loader)})]
+            all-paths (into [] cat [(:src paths) (:test paths) lib-paths])
+            class-loader-delay (delay (common/create-class-loader all-paths color-mode))]
         (when is-verbose (println (str "# paths:\n" all-paths "\n")))
         (doseq [current-test-runner test-runners-seeing-test-sources]
-          (when-not (setup-failed?)
+          (let [process-ns (when (test-runner-verifiers/valid-external-test-runner? current-test-runner)
+                             (test-runner-contract/external-process-namespace current-test-runner))
```
imrekoszo, line 145: question: if I'm reading the code right, if this ends up being non-null, then we're in external mode. For a test runner wanting to support both in-process and external modes, the way to do it would be to provide separate `create` functions for the two modes, is that correct?

imrekoszo:
Question to all of you: would it make sense to port over some of the code that deals with launching the external process from Sean's runner implementation?

Perhaps it's a little early for that. I'm thinking, it wouldn't be impossible to create a completely separate project (or even something inside the poly tool), which could implement the extra protocol added in this PR and "externalize" any in-process test runner, something along the lines of what Sean and I were [debating on slack](https://clojurians.slack.com/archives/C013B7MQHJQ/p1668195092811809?thread_ts=1668190791.646639&cid=C013B7MQHJQ).

components/workspace/src/polylith/clj/core/workspace/settings.clj
furkan3ayraktar: @tengstrand These are the two places in question:

[polylith.clj.core.workspace.settings](https://github.com/polyfy/polylith/blob/0628224b4e29671f39fc41463b377675e2a5b63b/components/workspace/src/polylith/clj/core/workspace/settings.clj)
[polylith.clj.core.workspace-clj.project-settings](https://github.com/polyfy/polylith/blob/0628224b4e29671f39fc41463b377675e2a5b63b/components/workspace-clj/src/polylith/clj/core/workspace_clj/project_settings.clj)

components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
furkan3ayraktar: @imrekoszo I've added more examples now. Let me know what you think.

components/test-runner-orchestrator/src/polylith/clj/core/test_runner_orchestrator/core.clj
furkan3ayraktar: A test runner can either be an in-process test runner and only reify the `TestRunner` protocol or an external test runner and reify both `TestRunner` and `ExternalTestRunner` protocols. It needs to return a valid external process namespace for it to be considered an external test runner.

furkan3ayraktar:
> Question to all: would it make sense to port over some of the code that deals with launching the external process from Sean's runner implementation?

This would remove the boilerplate needed for handling the Java subprocesses and running the setup/teardown functions. However, I don't know if it's important for an external test runner to control the process of creating the "external environment" needed to run the tests. I don't think I have enough knowledge to argue about this topic. What do you think, @seancorfield? Do you think it's valuable to control the creation of the Java subprocesses, or should we hide it inside the poly?

components/test-runner-contract/src/polylith/clj/core/test_runner_contract/interface.clj
```diff
@@ -66,14 +66,42 @@
...
+
+  Adding a test runner definition to a project will override the global test
+  runner. The project-a will use the global test runner, `my.namespace/create`
+  whereas project-b will use the default test runner.
+
+  {:test {:create-test-runner my.namespace/create}
+
+   :projects {\"project-a\" {:alias \"a\"}
+              \"project-b\" {:alias \"b\"
+                             :test {:create-test-runner :default}}}}"
```
imrekoszo: Reads very well in my opinion!

furkan3ayraktar: Thank you!

seancorfield:
> This would remove the boilerplate needed for handling the Java subprocesses and running the setup/teardown functions. However, I don't know if it's important for an external test runner to control the process of creating the "external environment" needed to run the tests. I don't think I have enough knowledge to argue about this topic. What do you think, @seancorfield? Do you think it's valuable to control the creation of the Java subprocesses, or should we hide it inside the poly?

I don't think we have enough knowledge/experience with this yet to decide to put the process spawning inside `poly` itself. As my code stands, there are several "knobs" that users might need to tweak. It doesn't support `JAVA_CMD` or a "standard" way to find the `java` command; I'm not very happy with the handling of JVM options (and that's critical to our test environment); it hasn't been tested on Windows (which is notorious for filename/path issues).

I _do_ feel that there's a fair bit of boilerplate/duplication between the, now three, different test runners in the `create` functions, so that might be an area that is worth looking at for pushing more of that into `poly`. Even there tho', there's already a TODO comment about narrowing what gets tested and it's possible that code might need to change between test runners. For example, it might be nice for us to have a test runner that is better integrated into our codebase at work and can make more assumptions about the environment in which it runs (for example, I might want to put the JVM options in an alias in `deps.edn` and read that via t.d.a when the runner is `create`'d).

There's also a use case that @imrekoszo raised in an earlier discussion: the test runner might prefer to be passed the classpath as an argument and build its own classloader for running tests (instead of running the subprocess directly via `java -cp ...`). We very specifically do not want that at work, but I can see why other test runners _might_ want to do that.

I can imagine someone deciding to build an external test runner based on Cognitect's `test-runner` which has a completely different API in terms of command-line arguments.

In all these cases, the actual subprocess is responsible for running the `setup-fn` and `teardown-fn`, regardless of what else the subprocess does, which means those need to be evaluated in the same context as the tests themselves.

TL;DR: adopt a "wait and see" approach to see what the community does with this new found superpower 😄

components/workspace/src/polylith/clj/core/workspace/settings.clj
furkan3ayraktar: @tengstrand and I discussed this and decided to move all test runner-related enrichments to the workspace map from the `workspace-clj` component to the `workspace` component.

You can find my changes here: https://github.com/polyfy/polylith/pull/263/commits/42d327e8211f0723d8c1f3cea3a4d03c847ccc9b

Let me know what you think!

imrekoszo (approved):
LGTM, thank you! polylith-kaocha tests are passing with 42d327e8211f0723d8c1f3cea3a4d03c847ccc9b

seancorfield:
I updated our repo at work to use the latest SHA for this PR (42d327e) and it seems to work, with the external test runner.

seancorfield:
I have updated https://github.com/seancorfield/polylith-external-test-runner to improve how it finds the `java` command to run the subprocess for the tests and also to support specifying JVM opts as an alias, which makes it a lot easier to pass multiple JVM options to the test subprocess. See the updated README, the workspace-level `deps.edn` file, and the newly-added tests.

seancorfield:
We've been running against this fork for a week now at work and it's working well, both locally and in CI, and has allowed us to reduce our BitBucket Pipelines CI configuration from `2x` back to the default size which has halved the cost of CI for us!

This PR also fixes #252 (memory consumption in large projects) and #253 (classloader isolation leaks).

Are there any obstacles to getting this merged at this point?

PR #269, opened 2022-12-13
fixes #268 switch all tools.deps.alpha references to non-alpha

### LINKED ISSUE #268: Switch to non-alpha tools.deps
`org.clojure/tools.deps {:mvn/version "0.16.1264"}` is available -- without "alpha".

I'll try to find time to put a PR together for this today.

seancorfield:
There was a mix of t.d.a versions and different sizes so I aligned them all.

If you'd rather I only changed the "active" t.d.a references, I can do another, smaller PR.


tengstrand (approved):
Looks great! Thank you for this!

PR #325, opened 2023-08-18
bb doc: improve docker dameon error msg

lread:
If the docker daemon is not running we now present a much more concise error message (the cause was previously shown but obscured by a big stack trace).


lread:
image

tengstrand:
Thanks!

lread:
You are most welcome! Don't forget to merge it!

PR #327, opened 2023-09-07
Review and update README

lread:
As a noob, I did not at first understand Polylith was separate from the Clojure poly tool and documented separately. I made this clearer in the README.
- changed the title to include "for Clojure"
- I used the term "conceptual" instead of "high-level" for Polylith gitbook docs.

I got very confused by the diagram.
It is very cool but I thought it deserved at least a high level description. I might have got it right, please do review.

Was not shy about using the word Polylith.
In a README, I felt it deserved to be repeated often.

Other changes were minor, like splitting up long sentences.

PR #329, opened 2023-09-09
First pass at adoc one sentence per line

lread:
As discussed on Slack.
See: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

Checked each file via GitHub rich diff and it found no visible changes.

I expect I missed a sentence or two, but can pick those up during content review.

Readme is handled separately in #327


tengstrand (approved):
Looks good!

PR #330, opened 2023-09-09
Review and update Introduction

lread:
Moved link to cljdoc to a tip that is only displayed if not already reading on cljdoc. Cljoc link adjusted to point to article we are viewing (introduction) instead of readme.

Same content but re-organized.

Starts with goal, then show the details to further entice reader.

Used bullet points to make text easier to digest.

Tried to be more succinct by getting rid of unnecessary words or rephrasing to use less words.

### DIFF
--- a/doc/introduction.adoc
+++ b/doc/introduction.adoc
@@ -5,23 +5,15 @@ ifndef::env-cljdoc[]
 TIP: For a richer experience, we recommend reading on {cljdoc-doc-url}/introduction[cljdoc].
 endif::[]
 
-This documentation offers a practical guide to using the Polylith tool.
-Our goal is to teach you the value of components and how they bring context to the development experience.
-We want to empower you to build decoupled and scalable systems from day one.
+This documentation aims to be a practical guide to this tool with lots of code examples.
+We encourage you to follow the code examples and try it out yourself.
+We will guide you through the steps of creating a workspace with projects composed of components, bases, and libraries and how to work with them from the development environment and the interactive shell.
 
-We have included many code examples.
-We strongly encourage you to try them out!
+We will give a short introduction to tools.deps and how to use tools.build to create deployable artifacts.
+We will explain how git is used to tag the code and how it enables us to test and release the code incrementally.
 
-You will learn how to:
+We will show how profiles will help us work from a single development environment for maximum efficiency and how dependencies and library usage can be displayed.
 
-* Create a workspace with projects composed of components, bases, and libraries.
-* Use the development environment and the interactive shell.
-* Use tools.deps and then tools.build to create deployable artifacts.
-
-We will explain how:
-
-* We leverage git tagging to support incremental tests and releases.
-* Profiles allow us to work efficiently from a single development environment.
-* To easily understand your dependency and library usage.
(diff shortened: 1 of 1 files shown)


lread:
@tengstrand my proposal for changes to the Introduction. 
Lemme know what you think.

doc/introduction.adoc
```diff
@@ -1,15 +1,27 @@
 = Introduction
 :cljdoc-doc-url: https://cljdoc.org/d/polylith/clj-poly/CURRENT/doc
 
-This {cljdoc-doc-url}/readme[documentation] aims to be a practical guide to this tool with lots of code examples.
-We encourage you to follow the code examples and try it out yourself.
-We will guide you through the steps of creating a workspace with projects composed of components, bases, and libraries and how to work with them from the development environment and the interactive shell.
+ifndef::env-cljdoc[]
+TIP: For a richer experience, we recommend reading on {cljdoc-doc-url}/introduction[cljdoc].
```
tengstrand, line 5: I really like this change!

doc/introduction.adoc
```diff
@@ -1,15 +1,27 @@
...
+* Use the development environment and the interactive shell.
+* Use tools.deps and then tools.build to create deployable artifacts.
+
+We will explain how:
+
+* We leverage git tagging to support incremental tests and releases.
+* Profiles allow us to work efficiently from a single development environment.
+* To easily understand your dependency and library usage.
 
 Let's begin!
```
tengstrand: Hi there!

I didn't give any boundaries on how to change the text, and that's one of the reasons you started with one page at a time to adjust this while we go through the documentation.

Both versions are good English, and I double checked that with a non-technical native English friend, that also work as an English teacher. The only feedback on the two versions she gave was that it should be an s in "one of the hardest and most important things" (which was not introduced by you). Other than that, she thought both versions were equally good.

The problem here, as I see it, is that the sound of the text is completely changed.
When the documentation gets more technical, it's less important to keep the voice of the text through out the documentation, but here I think we should. 

So if something is bad English, unclear, in the wrong order, or hard to understand, then change it. 
Almost as if this was a book, and your task was to make the text more understandable, without completely making the text your own.

It's also important that you're a developer trying to understand Polylith and the poly tool, so I couldn't have a non-technical person doing the job either, whether it happens to be an English teacher!

So to sum things up, keep my original text, add the tip and an s!

I hope you are still interested in continuing!

tengstrand: And one more thing. I wanted the introduction to be in this style, like someone telling a story, and not bullet points.

lread:
Due to some GitHub glitch, @tengstrand's PR feedback never made it to this PR.
But he shared the following on Slack:

image

image

image

lread:
Yeah, that's fine. Can do. Not trying to make the text my own, but clearer and easier to digest, but these things are subjective, so I'm happy to tweak.

lread:
@tengstrand re:

> The only feedback on the two versions she gave me was that it should be an s in "one of the hardest and most important things".

I don't see this text.

tengstrand (approved):
A lot of discussions on this one! Let's continue with the other pages!

PR #331, opened 2023-09-13
Review and update Install doc

lread:
Formatted overview on ways to install to a table.
For me this works better to compare variations.
It also helped me to fill in some missing blanks.
And to remove some details unnecessary for an overview (I moved some of these details down to specific install sections).
Let me know if this works for you too.

Noticed that overview mentions how to switch versions but install instructions do not talk about this. Did nothing about this yet.

Subtle: tried to use "we" only when talking from Polylith core team, and used "you" when talking to and guiding the reader.

Assumed poly core team preference is poly as stand-alone and/or Clojure CLI dep. Can adjust advice if this is not the case.
But making some clear suggestion avoids overwhelming newcomers.

Noticed some prerequisite and post-install advice repeated in some install instructions. But inconsistently. I moved these to adoc custom attributes so we have less copy/paste.

Now using adoc `subs="attributes+"` option in code blocks. This way code blocks can now reflect the `:poly-version:` attribute.

Moved clojure cli dep install instructions before clojure tools install instructions. This matches the overview order, and it (maybe?) the more likely order of interest for readers.

Added "stand-alone" to all headings that describe stand-alone installation. Just to make it extra clear.

Moved logging deps config to tools deps usage section as it only applies to that usage.

Re-orged Clojure Tools instructions a bit.
Clarified, I think, how it processes args.

Noticed that JVM options only applies to stand-alone variants to added that detail. I also noticed that it does not give an example, which I thought would be useful, but did nothing about that.

### DIFF
--- a/doc/install.adoc
+++ b/doc/install.adoc
@@ -3,59 +3,80 @@
 :poly-version: 0.2.18
 :cljdoc-api-url: https://cljdoc.org/d/polylith/clj-poly/CURRENT/api
 
-There are several ways to install the `poly` tool, which determines how the `poly` tool will be executed:
+There are 3 ways to install the `poly` tool:
 
-1. As a stand-alone command line tool, executed as a Java program, e.g. `poly help`:
-* The `poly` tool can be executed from any directory.
-* A https://github.com/polyfy/polylith/blob/master/build/resources/brew/exec[script] named `poly` is installed on your local disk and added to the system path, with a content like this: `java -jar poly.jar`
+[cols="20,27,27,27"]
+|===
+| | 1. Stand-alone | 2. link:#use-as-dependency[Clojure CLI dep] | 3. link:#install-as-clojure-cli-tool[Clojure CLI tool]
 
-2. By the https://clojure.org/guides/deps_and_cli[Clojure CLI], e.g. `clojure -M:poly help`:
-* The `poly` tool can only be executed from inside a Polylith xref:workspace.adoc[workspace] directory (it uses the `:poly` alias in `./deps.edn`).
-* Which version of the `poly` tool to use can easily be selected by changing the `sha` in `./deps.edn`.
+| Example usage
+| `poly help`
+| `clojure -M:poly help`
+a| `clojure -Tpoly help` +
+⚠️ xref:clojure-cli-tool.adoc[command line uses Clojure Tools exec args syntax] 
 
-3. As a https://clojure.org/reference/deps_and_cli#tool_install[Clojure CLI tool], e.g. `clojure -Tpoly help`:
-* The `poly` tool can be executed from any directory.
-* Different versions of the `poly` tool can easily be installed and switched to.
-* It doesn't follow the `poly` way of passing in arguments, which is described xref:clojure-cli-tool.adoc[here].
+| Installed via
+| Homebrew on macOS, otherwise via Nix or a manual download
+| Adding a `:poly` alias in `./deps.edn` 
+| Clojure Tools
 
-With this said, it's useful to install the tool both as a stand-alone command line tool, and any or both of the other two variants, but it's mainly up to your preferences how you prefer to use the tool.
+| Can be invoked from
+| Any directory 
+| A Polylith xref:workspace.adoc[workspace] directory 
+| Any directory 
 
-Now it's time to install the `poly` tool, so please follow the installation instructions for your operating system of choice.
+| Switch `poly` version via
+| `@<version>` if using HomeBrew 
+| Updating dependency for `:poly` alias in `./deps.edn` 
+| Reinstalling with Clojure Tools
+
+|===
+
+We prefer to install `poly` as a stand-alone and/or as a Clojure CLI dep.
+Your preference is fine too; there are no wrong choices.
+
+:prerequisites: First, ensure that you have https://clojure.org/guides/getting_started[Clojure CLI] and https://git-scm.com/book/en/v2/Getting-Started-Installing-Git[git] installed. +
+If you are installing git for the first time, don't forget to configure your git https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git[user.name] and https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-in-git[user.email].
+
+:post-install: pass:n[You will soon create your first xref:workspace.adoc[workspace] - for now, only `version`, `help`, `doc`, and `create workspace` commands will work.]
 
 [#install-on-mac]
-== Install on Mac
+== Install as Stand-alone on macOS
 
-To use the `poly` tool and to get access to all the features in tools.deps, make sure you have https://clojure.org/guides/getting_started[CLI tools] and https://git-scm.com/book/en/v2/Getting-Started-Installing-Git[git] installed.
-If you install git for the first time, don't forget to set the https://docs.github.com/en/github/using-git/setting-your-username-in-git[user] name and email.
+{prerequisites}
 
-To install the `poly` tool on Mac, execute:
+To install the `poly` tool on macOS, run:
 
 [source,shell]
 ----
 brew install polyfy/polylith/poly
 ----
 
-If you get the error "openjdk-13.0.2.jdk could not be opened...", do this:
+The `poly` https://github.com/polyfy/polylith/blob/master/build/resources/brew/exec[script] is now installed and added to the system path. 
+The script uses Java to launch the poly tool.
+
+If you get an error such as "openjdk-13.0.2.jdk could not be opened...", try:
 
 * Open MacOS `System Preferences > Security & Privacy > General`.
 * Click Allow at the bottom for `openjdk-13.0.2.jdk`.
 * Run `brew install polyfy/polylith/poly` again.
 
-Verify the installation by executing `poly help`.
+Verify the installation by running `poly help`.
+
+{post-install}
 
 [#install-on-linux]
-== Install on Linux
+== Install as Stand-alone on Linux
 
-To use the `poly` tool and to get access to all the features in tools.deps, make sure you have https://clojure.org/guides/getting_started[CLI tools] and https://git-scm.com/book/en/v2/Getting-Started-Installing-Git[git] installed.
-If you install git for the first time, don't forget to set the https://docs.github.com/en/github/using-git/setting-your-username-in-git[user] name and email.
+{prerequisites}
 
-To install the `poly` command on Linux:
+To install the `poly` tool on Linux:
 
 * Download the https://github.com/polyfy/polylith/releases/latest[latest release] of the poly jar, e.g.: `poly-{poly-version}.jar`.
-* Create a directory, e.g. `/usr/local/polylith`, and copy the jar file to that directory.
-* Create a file with the name _poly_ and put it in e.g. `/usr/local/bin` with this content:
-
-[source,shell]
+* Create a directory, e.g., `/usr/local/polylith`, and copy the jar file to that directory.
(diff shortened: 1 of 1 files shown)


lread:
@tengstrand This one took some time to review. 
Let me know what you think, we can tweak it as needed.

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
 
-[source,shell]
-----
-clojure -Ttools install io.github.polyfy/polylith '{:git/tag "v0.2.18" :deps/root "projects/poly"}' :as poly
-----
+[#use-as-dependency]
+== Use as dependency
 
-An alternative is to give a `sha`, which allows us to use old versions or versions that hasn't been released yet, e.g.:
+An alternative way of executing the `poly` tool is to specify it as a dependency in your `./deps.edn` file.
```
tengstrand, line 186: Most sentences use `our` instead of `your`, or `we` instead of `you`, as if we did this together. Change this to make it consistent.

doc/install.adoc
```diff
@@ -3,59 +3,80 @@
...
+Your preference is fine too; there are no wrong choices.
+
+:prerequisites: First, ensure that you have https://clojure.org/guides/getting_started[Clojure CLI] and https://git-scm.com/book/en/v2/Getting-Started-Installing-Git[git] installed. +
+If you are installing git for the first time, don't forget to configure your git https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git[user.name] and https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-in-git[user.email].
+
+:post-install: pass:n[You will soon create your first xref:workspace.adoc[workspace] - for now, only `version`, `help`, `doc`, and `create workspace` commands will work.]
 
 [#install-on-mac]
-== Install on Mac
+== Install as Stand-alone on macOS
```
tengstrand, line 44: Do we need to use upper case for Stand-alone? I prefer stand-alone, if you don't have any strong opinions.

doc/install.adoc
```diff
@@ -3,59 +3,80 @@
...
-* The `poly` tool can be executed from any directory.
-* A https://github.com/polyfy/polylith/blob/master/build/resources/brew/exec[script] named `poly` is installed on your local disk and added to the system path, with a content like this: `java -jar poly.jar`
+[cols="20,27,27,27"]
+|===
+| | 1. Stand-alone | 2. link:#use-as-dependency[Clojure CLI dep] | 3. link:#install-as-clojure-cli-tool[Clojure CLI tool]
 
-2. By the https://clojure.org/guides/deps_and_cli[Clojure CLI], e.g. `clojure -M:poly help`:
-* The `poly` tool can only be executed from inside a Polylith xref:workspace.adoc[workspace] directory (it uses the `:poly` alias in `./deps.edn`).
-* Which version of the `poly` tool to use can easily be selected by changing the `sha` in `./deps.edn`.
+| Example usage
```
tengstrand: The text in this column (for all rows in the table) could be bold (to match the top headers) and end with a `:` e.g. **Example usage:**

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
 
-Installed versions can be listed with:
+Replace `INSERT-LATEST-SHA-HERE` with a commit SHA from the Polylith repository, e.g. the https://github.com/polyfy/polylith/commits/master[latest] from the `master` branch.
+
+After adding one of the `:poly` aliases as described above, test the installation via `clojure -M:poly help`.
 
-[source,shell]
+{post-install}
+
+=== Testing from Master
```
tengstrand: Change to "Run from master" or similar, because many users use the latest commit in master to get access to the latest functionality (not only for testing). Maybe Master can be in lower case, because the name of that branch is `master`?

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
+
+[source,shell,subs="attributes+"]
 ----
-clj -Ttools list
+clojure -Ttools install io.github.polyfy/polylith '{:git/tag "v{poly-version}" :deps/root "projects/poly"}' :as poly
 ----
 
-We can get basic built-in help via the CLI's help machinery:
+An alternative to specifying a `:git/tag` is to specify a `:git/sha`.
+This allows you to install any git revision poly and is often used to test out a version that has not been officially released yet:
```
tengstrand, line 267: I try to use `us` (and we) when possible instead of `you`, as if we are doing this together.

doc/install.adoc
```diff
@@ -3,59 +3,80 @@
...
+
+|===
+
+We prefer to install `poly` as a stand-alone and/or as a Clojure CLI dep.
+Your preference is fine too; there are no wrong choices.
+
+:prerequisites: First, ensure that you have https://clojure.org/guides/getting_started[Clojure CLI] and https://git-scm.com/book/en/v2/Getting-Started-Installing-Git[git] installed. +
+If you are installing git for the first time, don't forget to configure your git https://docs.github.com/en/get-started/getting-started-with-git/setting-your-username-in-git[user.name] and https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-personal-account-on-github/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-in-git[user.email].
+
+:post-install: pass:n[You will soon create your first xref:workspace.adoc[workspace] - for now, only `version`, `help`, `doc`, and `create workspace` commands will work.]
```
tengstrand, line 41: Change to "We will soon create our first...". I try to use `we` instead of `you` when possible, to emphasise that we are doing this together! (check the other occurrencies too)

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
 
-An alternative way of executing the `poly` tool is to specify it as a dependency, by giving a commit SHA, or a release version.
-To use it this way, add one of the following aliases to the `:aliases` section in your `./deps.edn`:
+=== Clojure Tool CLI Command Line Syntax
 
-Via Clojars:
+The command-line argument syntax for Clojure CLI tools uses an "exec args" format, essentially Clojure keyword/value pairs.
 
-[source,clojure]
+For this reason, it can be convenient to first launch a `poly` shell so you can use the conventional `poly` command-line argument syntax.
```
tengstrand, line 284: `we` instead of `you`.

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
+Let's say you want to get info on lines of code.
```
tengstrand, line 285: `we` instead of `you`

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
- :aliases {:poly {:extra-deps {polylith/clj-poly {:mvn/version "0.2.18"}}
-                  :main-opts  ["-m" "polylith.clj.core.poly-cli.core"]}}
-...
-}
+clojure -Tpoly shell
 ----
 
-[#github-dependency]
-Via GitHub:
+After which you can specify standard Polylith arguments:
```
tengstrand, line 292: `we`

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
+[source,shell,subs="attributes+"]
+----
+                  _      _ + _   _
```
tengstrand, line 295: The ascii-art doesn't render nicely for me. Some characters are black, others are grey. One solution could be to change from `shell` to `text`. What is `attributes+` doing?

doc/install.adoc
```diff
@@ -148,138 +176,170 @@ nix-shell -p polylith
...
-[source,clojure]
-----
-{:aliases  {...
-            :poly {...
-                   :extra-deps {...
-                                org.apache.logging.log4j/log4j-api {:mvn/version "2.20."}
-                                org.apache.logging.log4j/log4j-slf4j-impl {:mvn/version "2.20.0"}}}}
-----
+You might want to specify more RAM to the `poly` tool and/or where the configuration file for logging is located.
+These can be conveyed via the `JVM_OPTS` environment variable which `poly` will passes along to the Java runtime at launch of poly stand-alone.
```
tengstrand, line 347: I think the logging example is relevant, for not-techie users (as me!).

tengstrand (changes requested):

doc/install.adoc
lread: No strong opinion at all, but capitalization for section titles should probably be consistent, whatever convention you choose.

doc/install.adoc
lread: Yes, from my (admittedly huge) commit comment:

> Subtle: tried to use "we" only when talking from Polylith core team, and used "you" when talking to and guiding the reader.

My personal feeling is using "our" and "we" is a bit confusing and makes me sometimes wonder what "I" should be doing. It's subtle. I just did a web search and found this:
https://bernoff.com/blog/the-pronoun-conundrum-we-vs-you-in-business-advice

This is subjective and will adhere to your preference, lemme know.

doc/install.adoc
lread: Sure, will do!

doc/install.adoc
lread: Good point, I did not mean to use "Testing" will switch as recommended.
See my previous question for section title capitalization.

Aside: I expect your repo was long created before https://www.bbc.com/news/technology-53050955, but some readers might be sensitive to the term master. Not making a suggestion here, just a comment.

doc/install.adoc
lread: See previous comment on us and we.

doc/install.adoc
lread: Good catch. Will try changing code style to text.

See the commit comment for `attributes+`. This allows adoc attributes to be replaced in code block, in this case {poly-version}.

doc/install.adoc
lread: See commit comment:

> Moved logging deps config to tools deps usage section as it only applies to that usage.

doc/install.adoc
lread: Maybe it is nice to capitalize headings "Like normal sentences"? I think that is fine too, lemme know.

doc/install.adoc
tengstrand: Hmm, interesting! This is important, so let me check this and come back to you!

doc/install.adoc
tengstrand: Yes, capitalizing the headings, but lower case elsewhere.

doc/install.adoc
tengstrand: Good point. The default is `main` when you create a new Polylith workspace. I think we kept it to reduce complexity, don't remember (we had a discussion in the Polylith team).

doc/install.adoc
tengstrand: Okay, cool.

doc/install.adoc
lread: So... to make sure I got it: for headings you'd like for example:

== The Dog and Cat are Friends

Instead of 

== The dog and cat are friends

(So Stand-alone would remain capitalized?)
I'm fine either way. Lemme know.

lread:
Ok @tengstrand: I've made a pass-through and addressed your feedback.

Known remaining items that I see:

- [ ] you vs we (lemme know when you've had a chance to discuss/decide)
- [ ] convention for heading capitalization (I still am not entirely clear what you want)

tengstrand (approved):
This looks really good and all changes make sense to me! I haven't check with my "English teacher friend" yet regarding the use of you/we, but I just decided to go with "your" style for now and use `you` as default. It's really hard to decide when to use `we` or `you`, so having `you` as default kind of makes sense because it's "never wrong" which `we` sometimes is. You can use `we` whenever that feels convenient, but let's figure this out later!

PR #334, opened 2023-09-15
Review and update Upgrade doc

lread:
Done:
1. Don't immediately presume macOS.
2. Add a little generic text, from which I linked to other useful docs.
3. Match same install variations from Install doc
4. Add toc to match Install doc
5. Separate Linux and Windows (the grouping was convenient for author but not the reader).

Not done:
1. The is no mention of SNAPSHOT builds. A reader might wonder how to upgrade (not sure if this is even do-able for brew and nix):
   1. from a final release to a SNAPSHOT release
   2. from a SNAPSHOT release to a latest SNAPSHOT release under same version

### DIFF
--- a/doc/upgrade.adoc
+++ b/doc/upgrade.adoc
@@ -52,7 +52,7 @@ nix-env -u polylith
 
 == As Dependency
 
-Update your `polylith/clj-poly` dependency in your `deps.edn` to the current version.
+Update your `polylith/clj-poly` dependency in your `./deps.edn` to the current version.
 
 == Clojure CLI Tool
 
--- a/doc/upgrade.adoc
+++ b/doc/upgrade.adoc
@@ -1,18 +1,59 @@
 = Upgrade
+:toc:
 :poly-version: 0.2.18
 
-If the `poly` tools is installed, it can be upgrade by executing:
+If the `poly` tool has been xref:install.adoc[installed], it can be upgraded.
+
+Polylith's releases are found and described on our link:https://github.com/polyfy/polylith/releases[releases page].
+We document our release strategy in detail xref:polylith-ci-setup.adoc#release[here].
+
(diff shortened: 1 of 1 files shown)


doc/upgrade.adoc
```diff
@@ -1,18 +1,59 @@
...
+
+Followed by:
+[source,shell]
+----
+nix-env -u polylith
+----
+
+== As Dependency
+
+Update your `polylith/clj-poly` dependency in your `deps.edn` to the current version.
```
tengstrand: I have a habit of writing `./deps.edn` if we refer to the deps.edn file at the workspace root (bricks and projects have their own).

tengstrand (changes requested):
Made one comment, but except for that, it looks great with many improvements!

tengstrand:
I think it's fine that this page only explains how to upgrade the stand-alone poly tool. I don't know yet how to upgrade a stand-alone pre-release of the poly tool, but perhaps we can revisit this page when we have made our first SNAPSHOT release.

doc/upgrade.adoc
lread: Ya, that's a good convention, will do!

lread:
> I think it's fine that this page only explains how to upgrade the stand-alone poly tool.

To make sure I understand, are you suggesting I remove sections?:
- As Dependency
- Clojure CLI Tool

> I don't know yet how to upgrade a stand-alone pre-release of the poly tool, but perhaps we can revisit this page when we have made our first SNAPSHOT release.

Sounds good.

tengstrand:
> To make sure I understand, are you suggesting I remove sections?:
No, I was unclear. It's fine as it is!

lread:
@tengstrand, good to merge, I think.

tengstrand (approved):

PR #335, opened 2023-09-18
Review and update Migrate docs

lread:
As discussed on Slack:
1. delete content describing lein-project migration, it is too old to be relevant anymore
2. move Migrate to way down in toc; it is not often used
3. merge migrate-workspace-v1-to-v2.adoc into migrate.adoc
4. add link from Upgrade docs to Migrate docs

Changes made to migrate v0.1 to v0.2 content:
1. tell reader up front in tip when they need to migrate
2. move info on what happens if you upgrade but don't migrate to an aside (to make this info easier to scan over and/or skip)
3. re-order advice and tell the user what they should be doing before a migration first
4. add headings so reader can more easily scan content
5. use numbered lists for steps the reader should carry out, this helps a reader not lose their place or accidentally skip a step
6. as always, try to use less words and reword for clarity


doc/migrate.adoc
```diff
@@ -1,12 +1,102 @@
 = Migrate
 
-== Leiningen
+TIP: Migration is unnecessary if you are using `poly` v0.2 or above.
```
tengstrand: I suggest that we rephrase this to something like "Migration is unnecessary if  you created your workspace with `poly` 0.2.0-alpha10 or above."

doc/migrate.adoc
```diff
@@ -1,12 +1,102 @@
 = Migrate
 
-== Leiningen
+TIP: Migration is unnecessary if you are using `poly` v0.2 or above. 
+Instead, see xref:upgrade.adoc[upgrade].
 
-The old https://github.com/tengstrand/lein-polylith[lein-polylith] tool has reached the end of its life and has been replaced by this tools.deps based version.
-If you have any old Leiningen based projects to migrate, please follow the instructions https://github.com/tengstrand/lein-polylith/blob/migrate/migrate/migrate.md[here].
+== From `poly` v0.1
```
tengstrand: I would be very surprised if not all users have migrated their projects already, so this instruction has almost no value to the users.

Now I think that it's probably better to add a Migrate section in upgrade.adoc, and point to the file in GitHub file instead (we can leave that file as it is, because no one will use it anyway). Then we can remove Migrate from the left menu also.

FYI:
Internally in the code, we have the key `:ws-type` in the workspace structure, that is set to `:toolsdeps1` if the the bricks doesn't store their own `deps.edn` files + doesn't have a `workspace.edn` file at the root. Otherwise it is set to `:toolsdeps2`.

tengstrand (changes requested):
I could have foreseen this, then you would have avoided going through migrate.adoc.

doc/migrate.adoc
lread: > I would be very surprised if not all users have migrated their projects already, so this instruction has almost no value to the users.

That's what you previously said about `lein-polylith`, but have you decided this is true of v0.1 as well?

lread:
> I could have foreseen this, then you would have avoided going through migrate.adoc.

Doh! I guess for those few who have not migrated from v0.1, the process will be clearer.

doc/migrate.adoc
lread: Yeah, that's better, will do.

doc/migrate.adoc
lread: > Now I think that it's probably better to add a Migrate section in upgrade.adoc, and point to the file in GitHub file instead (we can leave that file as it is, because no one will use it anyway). Then we can remove Migrate from the left menu also.

Another idea is to create an Appendix heading and stick the Migrate doc under that.
This way at least all your docs are in one place.

lread: > FYI:
> Internally in the code, we have the key `:ws-type` in the workspace structure, that is set to `:toolsdeps1` if the the bricks doesn't store their own `deps.edn` files + doesn't have a `workspace.edn` file at the root. Otherwise it is set to `:toolsdeps2`.

This means nothing to me... yet! But as I get deeper, I'm sure I'll understand.

tengstrand: I think both variants will work (only having the migrate page in github or to have a Migrate section). poly 0.2.0-alpha10 was released two years ago, so we can be quite sure that everyone has migrated. I have no strong opinion about this, so you can decide what to do, and we can always change our mind later.

lread:
Ok, @tengstrand, as discussed, this one is effectively now: delete migrate docs!
Ready for review and merge.

tengstrand (approved):

PR #336, opened 2023-09-18
Review and update Example Systems doc

lread:
Minor changes on this one.

Added a tip about the poly reports: to let the reader know what they are and that they will learn about them soon.


doc/example-systems.adoc
```diff
@@ -1,6 +1,9 @@
 = Example systems
 
-If you want to start by seeing how a full-blown system looks like in Polylith, then have a look at the systems listed here.
+If you are curious about what a full-blown system looks like in Polylith, look at the examples listed here.
+
+TIP: The reports on this page are all generated by the `poly` tool.
```
tengstrand: Almost true, but they are actually generated by the polyx tool which has its own page (polyx.adoc).

tengstrand (changes requested):

doc/example-systems.adoc
lread: Thanks, will update!

lread:
@tengstrand, this one is ready for re-review and merge.

PR #337, opened 2023-09-18
Review and update Production Systems doc

lread:
Reformatted text I missed in first pass to one-sentence-per-line.

Added tip on the poly reports and what they might be.

Strengthened request for companies using Polylith to make contact.


lread:
@tengstrand For this one, out of respect to authors, I left content I thought probably came directly from companies as-is. 
In no case were the spellos. And only subjective issues like missing commas or long sentences.

doc/production-systems.adoc
```diff
@@ -1,6 +1,9 @@
 = Production systems
 
-Here we list some of the companies that use Polylith in production.
+Here, we list some companies that use Polylith in production.
+
+TIP: The reports on this page are all generated by the `poly` tool.
```
tengstrand, line 5: Almost true, but they are actually generated by the `polyx` tool which has its own page (polyx.adoc).

doc/production-systems.adoc
```diff
@@ -1,6 +1,9 @@
...
+You'll learn all about them as you read through the `poly` tool documentation.
```
tengstrand, line 6: See previous note.

doc/production-systems.adoc
```diff
@@ -16,8 +19,7 @@ image::images/production-systems/funnel.png[link="images/production-systems/funn
 
 https://funnel.io/[Funnel] helps companies collect, prepare and analyze all their marketing data with ease.
 The company was founded in 2014, has 1000+ customers, and integrates with 500+ marketing apps and platforms.
-The tech stack differs from team to team.
-Python, Typescript, Rust, and Clojure are all languages in use.
+The tech stack differs from team to team. Python, Typescript, Rust, and Clojure are all languages in use.
 The Clojure adopters chose Polylith in order to get a smooth development experience and to be able to separate how services are deployed and run from how functionality is developed and re-used.
 
 == World Singles Networks
```
tengstrand, line 25: This section is written by Sean Corfield. He is busy, and has not time to look into your suggested changes, so I suggest that we instead revert changes for World Singles Networks. Only if we find errors or follow the rule of keeping sentences on one line, we should change.

tengstrand (changes requested):

doc/production-systems.adoc
lread: See my commit comment:
> Reformatted text I missed in first pass to one-sentence-per-line.

And my PR comment:
> @tengstrand For this one, out of respect to authors, I left content I thought probably came directly from companies as-is.
In no case were the spellos. And only subjective issues like missing commas or long sentences.

Sooo... there are no content changes here!

doc/production-systems.adoc
lread: I'll likely change this to "read through our documentation" to avoid talking too much about polyx.

tengstrand (approved):
Looks great!

PR #339, opened 2023-09-21
Review and update Workspace doc

lread:
Reword:
1. Change "addicted" to "enamored". I like the intent of "addicted" but I think we can use something as powerful but more sensitive.

Clarifications:
1. mention examples assume reader is using `poly` installed as stand-alone
2. mention a git repo is, by default, created
3. mention that examples commit the created workspace
4. move `branch:` example to a tip, the reader will be less likely to think they also need to run this variation
5. add trailing forward slash to directories. It makes it easier to see that they are directories
6. add note that user will need to commit the generated workspace files themselves when creating from an existing git repo
7. mention what `deps.edn` aliases are used for.

Content:
1. Update generated deps.edn `:poly` alias to use `:mvn/version`
2. Remove example and text around fetching `:latest-sha`

### DIFF
--- a/doc/workspace.adoc
+++ b/doc/workspace.adoc
@@ -65,7 +65,7 @@ Each top-level directory contains a specific Polylith concept:
 
 This structure gives a consistent shape to all Polylith projects.
 The familiar convention ensures that all developers, be they new or veterans, will quickly get started on Polylith systems that are new to them.
-We think you will become enamored with the power and simplicity the Polylith structure gives to your projects!
+We think you will fall in love with the power and simplicity the Polylith structure gives to your projects!
 
 ****
 The `bases`, `components`, and `projects` directories include a `.keep` file.
--- a/doc/workspace.adoc
+++ b/doc/workspace.adoc
@@ -17,7 +17,7 @@ poly create workspace name:example top-ns:se.example :commit
 The `poly` tool has created an `example` workspace in the `main` branch of a new git repository and committed all files. 
 
 TIP: The `se.example` top namespace is just an example.
-Choose what makes sense for your Polylith workspace.
+Stick with this name when following this example, but choose whatever name makes sense when creating your own Polylith workspaces.
 
 [TIP]
 ====
--- a/doc/workspace.adoc
+++ b/doc/workspace.adoc
@@ -1,65 +1,81 @@
 = Workspace
+:poly-version: 0.2.18
 
-The workspace directory is the place where all our code and most of the xref:configuration.adoc[configuration] lives.
+The workspace directory is where all your code and most of your xref:configuration.adoc[configuration] lives.
 
-Let’s start by creating the example workspace with the top namespace `se.example` by using the xref:commands.adoc#create-workspace[create workspace] command.
+NOTE: Unless otherwise noted, our examples assume `poly` is xref:install.adoc[installed] as a stand-alone tool. 
+Adjust your command line if you've installed `poly` differently.
 
-Make sure you execute the command outside a git repository:
+Let's create an example workspace with a top namespace of `se.example` using the xref:commands.adoc#create-workspace[create workspace] command.
+By default, this command also creates a git repository, so execute it outside any existing git repository: 
 
 [source,shell]
 ----
 poly create workspace name:example top-ns:se.example :commit
 ----
 
-This will create a workspace in the `main` branch.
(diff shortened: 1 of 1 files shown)


doc/workspace.adoc
```diff
@@ -1,65 +1,81 @@
...
 ----
 poly create workspace name:example top-ns:se.example :commit
 ----
 
-This will create a workspace in the `main` branch.
-By giving `branch:BRANCH-NAME` the workspace can be created in a different branch, e.g.:
+The `poly` tool has created an `example` workspace in the `main` branch of a new git repository and committed all files. 
+
+TIP: The `se.example` top namespace is just an example.
+Choose what makes sense for your Polylith workspace.
```
tengstrand: Add something like "but if you want to follow this example here, you need to set it to 'se.example'". Except from that, I like the change.

doc/workspace.adoc
```diff
@@ -1,65 +1,81 @@
...
 
-The directory structure is designed for quick navigation and ease of use.
-It helps us to understand and find all our service-level building blocks, which lets us reason about the system at a higher level.
+We designed the directory structure for quick navigation and ease of use.
+Your service-level building blocks are easy to find and understand, which helps you to reason about your system at a higher level.
 
-Each top-level directory contains a specific type of Polylith concept:
+Each top-level directory contains a specific Polylith concept:
 
 * A xref:base.adoc[base] is a building block that exposes a public API to external systems.
```
tengstrand: We could change to something like "A base is a building block that exposes a public API to the outside world, e.g. external systems and users (if e.g. a command line tool).

doc/workspace.adoc
```diff
@@ -1,65 +1,81 @@
...
-This will create a workspace in the `main` branch.
-By giving `branch:BRANCH-NAME` the workspace can be created in a different branch, e.g.:
+The `poly` tool has created an `example` workspace in the `main` branch of a new git repository and committed all files. 
+
+TIP: The `se.example` top namespace is just an example.
+Choose what makes sense for your Polylith workspace.
+
+[TIP]
+====
+You can override the git branch with the `branch:` option.
```
tengstrand: Here it makes sense to mention that `main` is the default branch (if they are fine with `main` then can skip `branch:`).

doc/workspace.adoc
```diff
@@ -1,65 +1,81 @@
...
 * A xref:project.adoc[project] specifies a deployable artifact and what components and bases it contains.
 
-* The xref:development.adoc[development] project (`development` + `deps.edn`) is used to work with the code in one place.
+* The xref:development.adoc[development] project (`development/` + `deps.edn`) supports working on all code from one place.
 
-This structure gives a consistent shape to all Polylith projects and ensures that both new developers and veterans can quickly understand and start working with systems that are new to them.
-We think you will soon be addicted to the power and simplicity the Polylith structure gives to your projects!
+This structure gives a consistent shape to all Polylith projects.
+The familiar convention ensures that all developers, be they new or veterans, will quickly get started on Polylith systems that are new to them.
+We think you will become enamored with the power and simplicity the Polylith structure gives to your projects!
```
tengstrand: Enamored was not part of my active vocabulary, and that is my only worry, that non-native speakers will find this hard to understand. I know that amore means love in Spanish, so I could figure it out. Maybe "fall in love" or something else could address this problem. But otherwise I'm pretty fine that we use enamored here.

tengstrand (changes requested):

doc/workspace.adoc
lread: Good point! I'll reword it to make that clearer.

doc/workspace.adoc
lread: Sounds good, I'll leave out the part about the command line tool because I think that kind of detail can be covered in Base doc.

doc/workspace.adoc
lread: Yes, will clarify that

doc/workspace.adoc
lread: Good feedback. I wonder if "fall in love" is maybe too over the top? 
But it is more accessible and I can't think of anything else so let's go with that.

lread:
Ok, @tengstrand, all review feedback has been addressed!

I don't want to put more in this PR so will deal with describing how to bootstrap using `poly` as a dep separately.

tengstrand (approved):
Great improvements!

PR #340, opened 2023-09-21
Revise poly as a dependency workflow

lread:
The old workflow did not make sense.


tengstrand (approved):
Nice use of the the bootstrapping!

PR #342, opened 2023-09-23
Review and update Component doc

lread:
Because most IDEs now support :local/root :extra-deps, moved this content out to its own section so it can more easily be skipped.

Don't assume all users are using Cursive.
Describe Cursive specific tip as such.

Otherwise, clarity and readability changes.


doc/component.adoc
```diff
@@ -1,68 +1,73 @@
...
+The `:auto-add` option tells `poly` to automatically add files generated by the xref:commands.adoc#create[create] command to xref:git.adoc[git].
+This behavior will be helpful in the examples that follow. 
 
-If `:auto-add` is set to `false`, then we have to manually add created files to git, after the xref:commands.adoc#create[create] command has been executed, or pass in `:git-add` to explicitly add them, which will have the same effect as setting `:auto-add` to `true`.
+****
+When `:auto-add` is `false`, you have to either:
 
-== How to execute commands
+* Manually `git add` files generated by xref:commands.adoc#create[create] commands yourself. 
+* Or use the `:git-add` option with `poly` `create` commands.
```
tengstrand: I think this change made it harder to understand, and important information was removed, for example that the user has to explicitly pass in the `:git-add` parameter when using the create command to get the same effect as setting the parameter in the config. 
I'm unsure if "use the x option" is more clear also than talking about parameters to commands. Maybe "pass in" is not correct English, so that can maybe be improved.

doc/component.adoc
```diff
@@ -1,68 +1,73 @@
...
+3. `clojure -Tpoly help` - as a Clojure CLI Tool.
 
-3. `clojure -Tpoly help` - as a Clojure CLI Tool
+Which forms are available depends on how you xref:install.adoc[installed] `poly`.
 
 All the documentation refers to the first form, even though the other two are also valid.
-There is also a fourth way, and that is to execute the command from a xref:commands#shell[shell], which can be started with:
 
-1. `poly shell`
+There is also a fourth way to execute `poly` commands.
```
tengstrand: I think breaking up this short sentence makes the sentence flow less well, so I prefer "There is also a fourth way to execute `poly` commands, and that is to use the `poly` shell". But if you have a strong opinion about this, then keep it as it is.

doc/component.adoc
```diff
@@ -1,68 +1,73 @@
...
 1. `poly`
-
 2. `clojure -M:poly`
-
 3. `clojure -T:poly shell`
 
 image::images/component/start-a-shell.png[]
 
-From here it's enough to enter the name of the command, e.g.:
+From the poly shell, it's enough to enter the name of the command, e.g.:
```
tengstrand, line 49: If the text comes directly under an image, it's usually about that picture. But if it's about the next picture, I have added a line break, to make it more clear for the user that the text is about the next image.

doc/component.adoc
```diff
@@ -1,68 +1,73 @@
...
-
-This way of executing commands is the most convenient way in most cases, because it gives us instant feedback, history, and autocomplete.
+In most cases, the poly shell is the most convenient way to execute commands.
+It gives you instant feedback, history, and autocomplete.
 Feel free to use the xref:shell.adoc[shell] from now on.
-It will make you more productive and enhances the development experience!
+It will make you more productive and enhance your development experience!
 
-Now let's continue with our example by executing the xref:commands.adoc#create-component[create component] command:
+TIP: Use `exit` to leave the `poly` shell.
```
tengstrand: ...or press <cmd>+C or <cmd> + D.

doc/component.adoc
```diff
@@ -162,9 +149,9 @@ The component also has its own `deps.edn` file that looks like this:
                   :extra-deps {}}}}
 ----
 
-It specifies that it has a `src`, `resources` and `test` directory.
+It specifies a `src`, `resources`, and `test` directory.
 
-The component was created with a `resources` directory:
+The `create component` command created the user component `resources` directory:
```
tengstrand: I have been consistent throughout the documentation to always link to the command's help page, and here we should put the in on "create component".

doc/component.adoc
```diff
@@ -236,7 +225,38 @@ image::images/component/add-user-namespaces.png[width=350]
   (core/hello name))
 ----
 
-Here we delegate the incoming call to the implementing `core` namespace, which is the most common way of structuring components in Polylith.
+The `interface` delegates the incoming call to the implementing `core` namespace, which is the most common way of structuring components in Polylith.
+
+In this example, we had you put all your implementing code in one single namespace.
+As a codebase grows, more namespaces can be added to the component as needed.
+There is no rule that the implementing namespace must be named `core`, but we chose that convention for this example.
```
tengstrand: The `core` namespace is created by the tool and is the default name of the implementing namespace, so I suggest that we change this sentence slightly.

doc/component.adoc
```diff
@@ -236,7 +225,38 @@ image::images/component/add-user-namespaces.png[width=350]
...
+
+* It's more readable.
+
+* It's consistent with how xref:project.adoc[projects] are specified.
+
+* You won't have to duplicate the bricks library dependencies in your workspace `./deps.edn`.
+
+* You can add or remove the `resources` directory from a brick without having to remember to update your workspace `./deps.edn`.
+
+If you want to compare the difference, take a look at how `article` was elegantly added as a single https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/blob/5b6df23d63500a4540b75308379e06dfdeb8b767/deps.edn#L7[dependency] versus adding it as two https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/blob/e6f7f200bc46e4e2595e123947eec442ad91c9ab/deps.edn#L7-L8[paths] in the RealWorld example app.
```
tengstrand, line 259: I like that we move this section to the end, and the "elegantly added" change to the sentence!

tengstrand (changes requested):

doc/component.adoc
lread: Oh, I thought this was a big improvement and much clearer.
I don't (yet) see the information loss.

Yes, "pass in" is awkward when talking about command lines.

Is your confusion with "parameter" vs "option"?  I thought these are command-line options.

doc/component.adoc
lread: I'll tend toward shorter sentences with the idea that they are easier to digest.
But I have absolutely no strong opinion here, so will join.

doc/component.adoc
lread: Yes in another PR I mentioned I think we can fixup any image spacing issues in cljdoc itself so you would not need these compenstations.

doc/component.adoc
lread: Would that work in Windows too?

doc/component.adoc
lread: That's fine can do. I typically would be sure to link on the first mention, but linking on every mention is absolutely fine too.

doc/component.adoc
lread: To add that fact? Sure, can do.

doc/component.adoc
lread: Coolio, glad you like it.

doc/component.adoc
lread: To give you an idea, here's how the asciidoctor tool renders spacing here:
image

Compared with current cljdoc rendering:
image

If I were to have cljdoc match the asciidoctor tool vertical image spacing, I think your issue with vertical spacing would go away. Am I right?

doc/component.adoc
lread: Anyway, I want to make sure this is easy to understand, so tell me more, and we'll get there.

lread:
@tengstrand, thanks so much for your thoughtful review!

doc/component.adoc
lread: Also, are you reviewing in rich text? I find sometimes the adoc one-sentence-per-line can make things seem to flow less than they actually do.

lread: Oh. Found something even better. I'll go with the much more succinct:

A fourth way to execute `poly` commands is through the `poly` shell.

doc/component.adoc
lread: Maybe leave as is and describe other ways to exit in shell command docs?

Probably too much detail to describe <kbd>Ctrl</kbd> vs <kbd>Cmd</kbd> and then also describe what works on Windows.

lread: Oh there are docs on quit in shell.adoc. I'll link to that and we'll update those docs when we get there.

doc/component.adoc
tengstrand: Yes, then I have no issue.

lread: Cool, I'll try to make the change on cljdoc sometime soon.

doc/component.adoc
lread: How about: 
> Or specify the `:git-add` option on the `poly` `create` command.

Does that work better for you?

(create should also be a link).

tengstrand: Yes, that works!

doc/component.adoc
```diff
@@ -1,68 +1,72 @@
...
 
-3. `clojure -Tpoly help` - as a Clojure CLI Tool
+Which forms are available depends on how you xref:install.adoc[installed] `poly`.
 
 All the documentation refers to the first form, even though the other two are also valid.
-There is also a fourth way, and that is to execute the command from a xref:commands#shell[shell], which can be started with:
 
-1. `poly shell`
+A fourth way to execute `poly` commands is through the `poly` shell.
+You start the poly shell via the xref:commands.adoc#shell[shell] command.
```
tengstrand: Maybe we should end this sentence with a colon.

tengstrand (approved):

PR #343, opened 2023-09-23
Review and update Interface doc

lread:
Add section headers.

Edits for clarity.

Some of the tips at the end of the doc were very confusing to me. After asking for clarification on Slack, rewrote some of them.

### DIFF
--- a/doc/interface.adoc
+++ b/doc/interface.adoc
@@ -52,7 +52,7 @@ Let's pretend we have the interface namespace `user.interface` containing the fu
 Because we are free to edit the `interface.clj` file for both user and admin, they can get out of sync if we are not careful. 
 Luckily, the `poly` tool will help us keep them consistent and complain if they differ when we run the xref:commands.adoc#check[check], xref:commands.adoc#info[info] or xref:commands.adoc#test[test] commands:
 
-[%autowidth]
+[cols="25,25,50"]
 |===
 |Check |Example mismatch| To protect against
 
--- a/doc/interface.adoc
+++ b/doc/interface.adoc
@@ -165,8 +165,7 @@ This list of tips makes more sense when you have used Polylith for a while, so b
 * Consider sorting interface namespace functions in alphabetical order for easy lookup. 
 Order functions in implementation namespaces freely.
 
-* The interface can expose the entity's name, e.g., `sell [car]`. 
-The implementing function can expose specific usage via destructuring, e.g., `sell [{:keys [model type color]}]`.
+* The interface can expose the entity's name, e.g., `sell [car]`, while the implementing function can expose specific usage via destructuring, e.g., `sell [{:keys [model type color]}]`.
 
 * It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in an interface to delegate to a single arity function in the implementing namespace:
 +
--- a/doc/interface.adoc
+++ b/doc/interface.adoc
@@ -83,8 +83,7 @@ Luckily, the `poly` tool will help us keep them consistent and complain if they 
 We often choose to have a single `interface` namespace in a component, but dividing the interface into several sub-namespaces is possible.
 To do so, create an `interface` package (directory) with the name `interface` at the component root and then add the sub-namespaces.
 
-You can find an example in Polylith itself.
-The `util` component divides its https://github.com/polyfy/polylith/tree/master/components/util/src/polylith/clj/core/util/interface[interface] into several sub-namespaces:
+You can find an example in Polylith itself, where the `util` component divides its https://github.com/polyfy/polylith/tree/master/components/util/src/polylith/clj/core/util/interface[interface] into several sub-namespaces:
 
 [source,shell]
 ----
--- a/doc/interface.adoc
+++ b/doc/interface.adoc
@@ -176,17 +176,17 @@ If `some-call` is a function in one `foo` interface, it must be a function in al
 We have this restriction because functions are composable, but macros are not.
 
 * Polylith will always recognize `interface` and `ifc` as interface namespace names.
-By default, it will generate code using `interface` as the interface namespace name.
+By default, it will generate code using `interface` as the interface namespace name when you create a component.
 You can override this default via `:interface-ns` in `./workspace.edn`.
 Scenarios:
 ** You want to share code between Clojure and ClojureScript via `.cljc` source files.
 Since `interface` is a reserved word in ClojureScript, it will cause problems.
 In this case, you can either:
-*** set `:interface-ns` to `ifc`, `poly` will use `ifc` as the interface namespace name for all generated code
-*** or leave `interface` as the default and manually rename any generated `interface` to `ifc` for the components that will also run from ClojureScript.
+*** set `:interface-ns` to `ifc`, `poly` will use `ifc` as the interface namespace name for generated code when you create a component
+*** or leave `interface` as the default and override by specifying `interface:ifc` as an option to xref:commands.adoc#create-component[create component] for components that will also run from ClojureScript.
 ** You want to consume Clojure code from another language on the JVM, e.g., Kotlin, where `interface` is a reserved word.
 You could set `:interface-ns` to anything that won't conflict, for the sake of this example, let's say `api`. 
-The `poly` tool would now use `api` for the interface namespace name when it generates code but also recognize `interface` and `ifc` as interface names.
+The `poly` tool would now use `api` for the interface namespace name when it generates code when you create a component, but also recognize `interface` and `ifc` as interface names.
 
 
 
--- a/doc/interface.adoc
+++ b/doc/interface.adoc
@@ -146,7 +146,7 @@ Order functions in implementation namespaces freely.
 * The interface can expose the entity's name, e.g., `sell [car]`. 
 The implementing function can expose specific usage via destructuring, e.g., `sell [{:keys [model type color]}]`.
 
-* It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in an interface to delegate to a single arity function in the implementing namespace. 
+* It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in an interface to delegate to a single arity function in the implementing namespace:
 +
 [source,clojure]
 ----
@@ -155,7 +155,7 @@ The implementing function can expose specific usage via destructuring, e.g., `se
   ([a b] (foo a b nil)))
 ----
 
-* It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#variadic-functions[variadic functions] in an interface to delegate to function in the implementing namespace that accepts the variadic portion as a vector.
+* It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#variadic-functions[variadic functions] in an interface to delegate to function in the implementing namespace that accepts the variadic portion as a vector:
 +
 [source,clojure]
 ----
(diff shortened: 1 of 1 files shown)


doc/interface.adoc
```diff
@@ -45,16 +49,19 @@ To give an example, let's pretend we have the interface namespace `user.interfac
   ...
 ----
 
-Now we are free to edit the `interface.clj` file for both `user` and `admin`, which means they can get out of sync if we are not careful enough.
-Luckily, the Polylith tool will help us keep them consistent, and complain if they differ when we run the xref:commands.adoc#check[check], xref:commands.adoc#info[info] or xref:commands.adoc#test[test] commands.
+Because we are free to edit the `interface.clj` file for both user and admin, they can get out of sync if we are not careful. 
+Luckily, the Polylith tool will help us keep them consistent and complain if they differ when we run the xref:commands.adoc#check[check], xref:commands.adoc#info[info] or xref:commands.adoc#test[test] commands.
```
tengstrand: Her should use "the poly `tool`" (where poly is marked as code, which is also a convention through out the documentation). In the high-level doc we sometimes use "the Polylith tool" because that documentation is generic, but here we know that we talk about the `poly` tool. I missed to change this, so this was not introduced by you.

doc/interface.adoc
```diff
@@ -126,46 +134,61 @@ There is no magic here, just include the definitions you want, like this:
...
+* Consider sorting interface namespace functions in alphabetical order for easy lookup. 
+Order functions in implementation namespaces freely.
 
-* If we have a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in the interface, a simplification can sometimes be to have a single arity function in the implementing namespace that allows some parameters to be passed in as nil.
+* The interface can expose the entity's name, e.g., `sell [car]`. 
+The implementing function can expose specific usage via destructuring, e.g., `sell [{:keys [model type color]}]`.
 
-* If using http://clojure-doc.org/articles/language/functions.html#variadic-functions[variadic functions] in the interface, a simplification is to pass in what comes after `&` as a vector to the implementing function.
+* It sometimes makes sense for a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in an interface to delegate to a single arity function in the implementing namespace.
```
tengstrand: End the sentence with a colon (the same applies to the next point).

doc/interface.adoc
```diff
@@ -126,46 +134,61 @@ There is no magic here, just include the definitions you want, like this:
...
-Finally, the interface namespace name can be changed in `:interface-ns` in `./workspace.edn`.
-Here are a few reasons why we would want to do that:
+* If the interface `foo` occurs in xref:#one-interface-in-multiple-components[multiple components], you cannot define `some-call` as a function in one `foo` interface and a macro in other `foo` interfaces. 
+If `some-call` is a function in one `foo` interface, it must be a function in all `foo` interfaces. If `some-call` is a macro in one `foo` interface, it must be a macro in all `foo` interfaces.
+We have this restriction because functions are composable, but macros are not.
 
-* We want to share code between Clojure and ClojureScript via `cljc` files.
-Since `interface` is a reserved word in ClojureScript, it could otherwise cause problems.
+* Polylith will always recognize `interface` and `ifc` as interface namespace names.
+By default, it will generate code using `interface` as the interface namespace name.
```
tengstrand: Add: "when you create a component." in the end

doc/interface.adoc
```diff
@@ -126,46 +134,61 @@ There is no magic here, just include the definitions you want, like this:
...
+You can override this default via `:interface-ns` in `./workspace.edn`.
+Scenarios:
+** You want to share code between Clojure and ClojureScript via `.cljc` source files.
+Since `interface` is a reserved word in ClojureScript, it will cause problems.
+In this case, you can either:
+*** set `:interface-ns` to `ifc`, `poly` will use `ifc` as the interface namespace name for all generated code
+*** or leave `interface` as the default and manually rename any generated `interface` to `ifc` for the components that will also run from ClojureScript.
```
tengstrand: We don't have to manually change the namespace here. Instead, we can give `interface:ifc` to the `create component` command, and it we use `ifc` instead as the interface name.

tengstrand (changes requested):
Great improvements! They will definitely make the text easier to understand for users new to Polylith.

doc/interface.adoc
```diff
@@ -126,46 +134,61 @@ There is no magic here, just include the definitions you want, like this:
...
-The reason for this restriction is that functions are composable, but macros are not, which could otherwise cause problems.
+* Because Polylith only allows the code under `src` to call `interface` code, you can think of publicly declared implementation functions as protected (as in Java).
+Because these "protected" functions are technically public, you can test and debug them more easily.
+For example, when stopping at a breakpoint to evaluate a "protected" function, you don't need to use the special syntax you would need to access a private function.
 
-Finally, the interface namespace name can be changed in `:interface-ns` in `./workspace.edn`.
-Here are a few reasons why we would want to do that:
+* If the interface `foo` occurs in xref:#one-interface-in-multiple-components[multiple components], you cannot define `some-call` as a function in one `foo` interface and a macro in other `foo` interfaces. 
+If `some-call` is a function in one `foo` interface, it must be a function in all `foo` interfaces. If `some-call` is a macro in one `foo` interface, it must be a macro in all `foo` interfaces.
+We have this restriction because functions are composable, but macros are not.
```
tengstrand: Here somewhere, we could mention that the `poly` tool also checks that the signature must match exactly, for functions and macros. So if the parameter list is `[a b c]` for function `foo` it has to be defined as `[a b c]` in `foo` for all the implementing components. It means that `[b a c]` is not accepted. Type hints are also checked, e.g. if a function is defined as `[^String x y]`, we have define it as `[^String x y]` everywhere for that function. This reduces the risk that we enter the parameters in the wrong order.

doc/interface.adoc
lread: Yeah, that's a good convention. I'll update.

doc/interface.adoc
lread: Good to know, will add.

doc/interface.adoc
lread: These matching requirements are described above under the "Declaring the Same Interface in Multiple Components" section. I think it would be clearer if this whole tip was moved up to that section where it would be in context. 

I also like your explicit function signature examples, I'll work them in.
I expect I could make `defn` vs `defmacro` less wordy and clearer with examples.

doc/interface.adoc
lread: Can do

doc/interface.adoc
lread: Can do

doc/interface.adoc
```diff
@@ -45,16 +49,42 @@ To give an example, let's pretend we have the interface namespace `user.interfac
   ...
 ----
 
-Now we are free to edit the `interface.clj` file for both `user` and `admin`, which means they can get out of sync if we are not careful enough.
-Luckily, the Polylith tool will help us keep them consistent, and complain if they differ when we run the xref:commands.adoc#check[check], xref:commands.adoc#info[info] or xref:commands.adoc#test[test] commands.
+Because we are free to edit the `interface.clj` file for both user and admin, they can get out of sync if we are not careful. 
+Luckily, the `poly` tool will help us keep them consistent and complain if they differ when we run the xref:commands.adoc#check[check], xref:commands.adoc#info[info] or xref:commands.adoc#test[test] commands:
+
+[cols="25,25,50"]
```
tengstrand, line 55: I really like this change of introducing a table here!

doc/interface.adoc
```diff
@@ -45,16 +49,42 @@ To give an example, let's pretend we have the interface namespace `user.interfac
...
-interface into several sub namespaces:
+|===
+
+== Splitting an Interface into Sub-Namespaces
+
+We often choose to have a single `interface` namespace in a component, but dividing the interface into several sub-namespaces is possible.
+To do so, create an `interface` package (directory) with the name `interface` at the component root and then add the sub-namespaces.
+
+You can find an example in Polylith itself.
+The `util` component divides its https://github.com/polyfy/polylith/tree/master/components/util/src/polylith/clj/core/util/interface[interface] into several sub-namespaces:
```
tengstrand: Here I prefer one sentence, because they are so coupled to each other:
You can find an example in Polylith itself, where the `util` component divides its...

doc/interface.adoc
```diff
@@ -67,10 +97,11 @@ util
...
+Dividing an interface like this can be handy to group and sub-group the interface functions.
+One typical usage is to place https://clojure.org/about/spec[clojure specs] in its own `spec` sub-namespace.
+You can see an example of this in the https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app[RealWorld example app], where the `article` component has both an `article.interface` and an https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/blob/master/components/article/src/clojure/realworld/article/interface/spec.clj[`article.interface.spec`] sub-interface.
 
-It can then be used from e.g. the https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/blob/master/bases/rest-api/src/clojure/realworld/rest_api/handler.clj[handler] namespace in `rest-api`:
+Here's an example of the sub-interface usage from the https://github.com/furkan3ayraktar/clojure-polylith-realworld-example-app/blob/master/bases/rest-api/src/clojure/realworld/rest_api/handler.clj[RealWorld example app handler] namespace in its `rest-api`:
```
tengstrand: End the sentence with "base" (that links to the `base` page):
...in it's `rest-api` base:

doc/interface.adoc
```diff
@@ -126,46 +157,57 @@ There is no magic here, just include the definitions you want, like this:
...
-* The interface can expose the name of the entity, e.g. `sell [car]`, while the implementing function can do the destructuring, e.g. `sell [{:keys [model type color]}]` which sometimes can improve the readability.
+* The interface docstrings should focus on what problem each function/macro solves, while the implementation docstrings can focus on concrete details.
 
-* If we have a http://clojure-doc.org/articles/language/functions.html#multi-arity-functions[multi-arity function] in the interface, a simplification can sometimes be to have a single arity function in the implementing namespace that allows some parameters to be passed in as nil.
+* Consider sorting interface namespace functions in alphabetical order for easy lookup. 
+Order functions in implementation namespaces freely.
 
-* If using http://clojure-doc.org/articles/language/functions.html#variadic-functions[variadic functions] in the interface, a simplification is to pass in what comes after `&` as a vector to the implementing function.
+* The interface can expose the entity's name, e.g., `sell [car]`. 
+The implementing function can expose specific usage via destructuring, e.g., `sell [{:keys [model type color]}]`.
```
tengstrand: Here the comparison is lost when we split the sentence in two. I recommend to put back "while" again, to make it clearer that they have two different roles.

tengstrand (changes requested):

doc/interface.adoc
lread: Glad you like it too!

doc/interface.adoc
lread: Yeah, I think we'll continue to have a healthy back-and-forth on sentence length.
I see no problem with joining here.

doc/interface.adoc
lread: Can do

doc/interface.adoc
lread: I find it clearer with 2 sentences.
They are in the same paragraph, after all.
But I think we are on subjective grounds here maybe? So also fine with joining them.
If only we had an English teacher to chime in! :slightly_smiling_face:

lread: Hmm... maybe you are right. That does seem to read better!

tengstrand (approved):

PR #344, opened 2023-09-25
Let readers know doc improvements are a wip

lread:


readme.adoc
```diff
@@ -10,6 +10,8 @@ https://clojurians.slack.com/messages/C013B7MQHJQ[image:https://badgen.net/badge
 
 == Welcome to the wonderful world of Polylith for Clojure!
 
+NOTE: *Work in Progress*: While we are very proud of our documentation, we are currently working to make it even better!
```
tengstrand: Perhaps we should add information that this documentation is a SNAPSHOT build from the issue-318 branch (with a link to the polylith repository for that branch). Otherwise, it will not be obvious which code it corresponds to, when you run the cljdoc version of it.

We could also (maybe) mention that in the future, all SNAPSHOT releases will be built from the `master` branch, but right now it's from `issue-318`.

tengstrand (changes requested):

readme.adoc
lread: I was thinking keeping it vague was nice in that it would let you merge at any time without changing any text.

But, that's fine, I'll add in those details.

You'll just have to remember to update the text before the final merge to master!

tengstrand (approved):

PR #345, opened 2023-09-25
Add developer notes on documentation

lread:
This is not a review of the Developing Poly doc, just an addition of some notes on documentation.


tengstrand (approved):
Looks great!

PR #346, opened 2023-09-26
Review and update Base doc

lread:
Edits for clarity.

Distinguish cursive advice by making it a tip.

Experiment with AsciiDoc code callouts.
We'll backtrack and add them to previously reviewed docs if they seem to add value. (I think they do).

### DIFF
--- a/doc/base.adoc
+++ b/doc/base.adoc
@@ -8,8 +8,7 @@ A `base` is similar to a xref:component.adoc[component] except for two things:
 
 The lack of an xref:interface.adoc[interface] makes bases less composable than components.
 This is fine; the sole purpose of a `base` is to act as a bridge between the real world and the xref:component.adoc[components] it delegates to.
-A `base` gives us the modularity and structure we need to build simple, understandable services and tools.
-
+This structure gives us the modularity we need to build simple, understandable services and tools.
 
 Let's continue with our tutorial.
 The last thing you did was xref:component.adoc#create-component[create a component].
--- a/doc/base.adoc
+++ b/doc/base.adoc
@@ -1,29 +1,32 @@
 = Base
 
-A `base` is similar to a `component` except for two things:
+A `base` is similar to a xref:component.adoc[component] except for two things:
 
-* It doesn't have an `interface`.
+* It doesn't have an xref:interface.adoc[interface].
 
 * It exposes a public API to the outside world.
 
-The lack of an `interface` makes bases less composable compared to components.
-This is okay because they serve a different purpose which is to be a bridge between the real world and the components it delegates to.
-This gives us the modularity and structure we need to build simple and understandable services and tools.
+The lack of an xref:interface.adoc[interface] makes bases less composable than components.
+This is fine; the sole purpose of a `base` is to act as a bridge between the real world and the xref:component.adoc[component] it delegates to.
+A `base` gives us the modularity and structure we need to build simple, understandable services and tools.
 
-Let's create the `cli` base to see how it works, by executing the xref:commands.adoc#create-base[create base] command:
+
+Let's continue with our tutorial.
+The last thing you did was xref:component.adoc#create-component[create a component].
+Run the following xref:commands.adoc#create-base[create base] command to create a `cli` `base`:
 
 [source,shell]
 ----
 poly create base name:cli
 ----
 
-Our workspace should now look like this:
+Your workspace should now look like this:
 
 [source,shell]
 ----
 example
-├── bases
-│   └── cli
+├── bases # <1>
+│   └── cli # <2>
 │       ├── deps.edn
 │       ├── resources
 │       │   └── cli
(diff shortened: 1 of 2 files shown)


doc/base.adoc
```diff
@@ -1,29 +1,32 @@
...
-* It doesn't have an `interface`.
+* It doesn't have an xref:interface.adoc[interface].
 
 * It exposes a public API to the outside world.
 
-The lack of an `interface` makes bases less composable compared to components.
-This is okay because they serve a different purpose which is to be a bridge between the real world and the components it delegates to.
-This gives us the modularity and structure we need to build simple and understandable services and tools.
+The lack of an xref:interface.adoc[interface] makes bases less composable than components.
+This is fine; the sole purpose of a `base` is to act as a bridge between the real world and the xref:component.adoc[component] it delegates to.
```
tengstrand: Should be "...components it delegates to" (plural).

tengstrand (changes requested):

doc/base.adoc
tengstrand: "components it delegate to"

lread:
@tengstrand ready for merge? Or do you have more feedback?

doc/base.adoc
```diff
@@ -1,29 +1,32 @@
...
+* It doesn't have an xref:interface.adoc[interface].
 
 * It exposes a public API to the outside world.
 
-The lack of an `interface` makes bases less composable compared to components.
-This is okay because they serve a different purpose which is to be a bridge between the real world and the components it delegates to.
-This gives us the modularity and structure we need to build simple and understandable services and tools.
+The lack of an xref:interface.adoc[interface] makes bases less composable than components.
+This is fine; the sole purpose of a `base` is to act as a bridge between the real world and the xref:component.adoc[components] it delegates to.
+A `base` gives us the modularity and structure we need to build simple, understandable services and tools.
```
tengstrand: I prefer "This gives us", because it's the collaboration between bases and components (the delegation to them via interfaces) that gives us the modularity, not just the bases alone.

doc/base.adoc
```diff
@@ -1,29 +1,32 @@
...
 
-Let's create the `cli` base to see how it works, by executing the xref:commands.adoc#create-base[create base] command:
+
+Let's continue with our tutorial.
+The last thing you did was xref:component.adoc#create-component[create a component].
+Run the following xref:commands.adoc#create-base[create base] command to create a `cli` `base`:
```
tengstrand: I would prefer to only have `cli` marked as code, because that's the thing that is interesting in this context.

doc/base.adoc
```diff
@@ -63,40 +66,43 @@ example
 ├── readme.md
 └── workspace.edn
 ----
+<1> new `cli` base
 
-Now we need to update `./deps.edn` with our newly created base:
+Edit your workspace `./deps.edn` to add your newly created `base`:
```
tengstrand: I think we can use base instead of `base` here.

tengstrand (changes requested):

doc/base.adoc
lread: Ah right, my grammar checker said that the reader might not know what "This" refers to.
My guess was wrong!

So the reader does not have to guess, instead maybe?:
"This structure gives us the modularity we need to build simple, understandable services and tools"

doc/base.adoc
lread: Yeah, I'm confused about what should be marked as code and what should not.
Let's work out a convention on slack and document it.

doc/base.adoc
lread: See last comment

doc/base.adoc
```diff
@@ -1,29 +1,32 @@
...
-* It doesn't have an `interface`.
+* It doesn't have an xref:interface.adoc[interface].
 
 * It exposes a public API to the outside world.
 
-The lack of an `interface` makes bases less composable compared to components.
-This is okay because they serve a different purpose which is to be a bridge between the real world and the components it delegates to.
-This gives us the modularity and structure we need to build simple and understandable services and tools.
+The lack of an xref:interface.adoc[interface] makes bases less composable than components.
+This is fine; the sole purpose of a `base` is to act as a bridge between the real world and the xref:component.adoc[components] it delegates to.
```
tengstrand: After "it delegates to", add "via the component's interfaces".

tengstrand: Maybe in parenthesis.

doc/base.adoc
tengstrand: The convention could be that we don't mark the interface/component/base/project as code from the same page as the "entity", e.g. from base.adoc, we don't refer to base as `base`. An exception could be the first time we introduce it (`base` in this example) but that should be optional, and only used if we think it adds clarity.

lread: Ok, so for example, for the base doc, we format as `base` for the first usage.
Other key concepts like [interface](#) or [component](#) will be links, so we don't have to worry about those.

But... we might find that always linking is a bit distracting (and of little value) if over-used for the same term.
For example, if we are talking about [interface](#) a bunch we probably want to instead use code formatting for subsequent uses of `interface` if nearby the link.
Or if that is distracting, use plaintext interface for subsequent uses.

Concrete rules are easier to follow... but does that work for you?

lread: Actually, let's go with plaintext for nearby subsequent uses of, from the above example, interface. 

All sound good?

tengstrand (approved):
Looks great!

Issue #348, opened 2023-10-01
doc: minor update reminders


lread:
From chat on Slack with @tengstrand:

I have wip in progress that conflicts with these secondary reviews. Hence this reminder issue.

## Workspace

- [x]  Replace: All techniques above still apply, but you will instead, for example:
With: All techniques above still apply, but you will instead create a workspace like so:
(With the idea it will probably be easier to grok for folks whose first language is not English.)

## Component

- [x] General -> Specific for: 
> Joakim: On the component page, I think we can leave out :extra-deps from "IDE Support for :local/root" because it's only :local-root that the IDEs have problem with
> Lee: Technically true, I was just following the spirit of the original text. But I think this would be an improvement.
The content should would also need to be tweaked a tiny bit to state the general issue before moving on to the specific.

## Example, Production Systems

- [x]  In the tip at the top of example-systems, we could mention that the individual three diagrams in each image can be generated by the `info`,  `deps`, and `libs` commands that we will cover later.

## Dev docs

- [x] Mention InkScape is used for doc diagrams. 
Convention of saving .svg next to .png.
Use cross-platform fonts (I think installing Microsoft fonts should be enough).

- [x] Mention SNAPSHOT check behavior.

## Doc
 - [x] The link "Different versions" does not work.

## General
- [x] do not -> don't
We'll prefer don't in general. Might also add note to dev docs on docs.

- [x] We've decided on "argument" instead of "parameter". I'll swing back to documents I've already edited and make the change, otherwise I'll handle this as I progress sequentially through docs.
- [x]  Add (1), (2)...numbering (adoc code callouts) to the code examples in already reviewed pages, like near "Your workspace directory structure" and "Let’s listen to the helpful reminder" in component.adoc.

- [x] Describe the tutorial example in some detail very early in the docs.
Then we can link back to that description from later tips.

lread:
All done!

PR #351, opened 2023-10-02
Review and update Shell doc

lread:
Added more headings.

Added more hand-holding, explanations and terminal examples.

It is nice to learn how to quit early.
Moved section right after instructions on how to launch.

Added necessary pre-steps for folks who want to try out switch-ws examples.

Edits for clarity.


doc/component.adoc
```diff
@@ -10,12 +10,12 @@ But before you do that, open `workspace.edn` in your editor and set `:auto-add`
...
 
 The `:auto-add` option tells `poly` to automatically add files generated by the xref:commands.adoc#create[create] command to xref:git.adoc[git].
-This behavior will be helpful in the examples that follow. 
+This behavior will be helpful in the examples that follow.
 
 ****
 When `:auto-add` is `false`, you have to either:
 
-* Manually `git add` files generated by xref:commands.adoc#create[create] commands yourself. 
+* Manually `git add` files generated by xref:commands.adoc#create[create] commands yourself.
```
tengstrand: Maybe we could add a link here to https://git-scm.com/docs/git-add, so that it gets even more clear what we are talking about?

doc/shell.adoc
```diff
@@ -1,187 +1,361 @@
...
-The goal of the shell is to make us more productive and to have more fun!
+Not only does the `shell` make you more productive, but it's also fun to use!
+It takes full advantage of https://github.com/jline/jline3[jline3] under the hood, which supports macOS, Linux, FreeBSD, Solaris, and Windows.
 
-The shell uses https://github.com/jline/jline3[jline3] under the hood which supports the following operating systems:
-Mac OS X, Linux, FreeBSD, Solaris, and Windows.
+Let's explore some shell features.
 
-Let's start a shell by typing `poly` from the workspace root:
+TIP: Our examples assume you are exploring from our ongoing tutorial `example` xref:workspace.adoc[workspace].
```
tengstrand, line 9: I like that we add a tip here. We could also add a link to the final version of the example, and inform the user that we have it [here](https://github.com/polyfy/polylith/tree/issue-318b/examples/doc-example).

doc/shell.adoc
```diff
@@ -1,187 +1,361 @@
...
-If a workspace has been exported to a file using something like `poly ws out:ws.edn` then we can switch to that workspace.
-Here we use the https://github.com/seancorfield/usermanager-example/tree/polylith[usermanager] project as an example (make sure to switch to the `polylith` branch, after cloning the repository):
+=== Switching to an Exported Workspace
 
-image::images/shell/shell-switch-ws-file.png[width=600]
+You can switch to an exported workspace.
 
-The prompt name is picked from the `:name` attribute in the `ws.edn` file and is prefixed with `file:`
-to indicate that it has been imported from a file.
+NOTE: You can export a workspace to a `.edn` file, e.g. via `poly ws out:ws.edn`.
```
tengstrand: We use `ws.edn` almost everywhere in the examples. We could use `realworld.edn` as an example, just to give an example that the file can have any name.

doc/shell.adoc
```diff
@@ -1,187 +1,361 @@
...
+
+=== Switch at Launch
+
+You can launch a shell and immediately switch to a workspace:
+
+[source,text,subs="+attributes"]
 ----
-$ poly ws-dir:examples/doc-example
+$ poly ws-dir:../clojure-polylith-realworld-example-app
                   _      _ + _   _
```
tengstrand, line 195: As you said before, I think in Slack, we should replace these two ascii versions. I can do that after this page has been approved (I put that on my todo list).

doc/shell.adoc
```diff
@@ -1,187 +1,361 @@
...
+=== Workspace Switch Mechanics
+
+Knowing how the shell supports workspace switching can be helpful.
+It is pretty simple:
+
+* After switching to workspace dir `D`, the shell silently appends `ws-dir:D` to each command before it runs it.
+* After switching to workspace export file `F`, the shell silently appends `ws-file:F` to each command before it runs it.
 
-The autocomplete helps us execute different commands with correct parameters:
+== Autocomplete Examples
```
tengstrand, line 224: I really like this improvement, with examples added!

tengstrand (changes requested):
Nice work, and great improvements!

doc/component.adoc
lread: Yes, that would probably help. Will do!

doc/shell.adoc
lread: Maybe we should talk about the tutorial example in some detail very early in the docs.
Then we can link back to that description from later tips. This could be a todo in #348.

doc/shell.adoc
lread: Sounds good, will do!

doc/shell.adoc
lread: Yeah probably. I guess it really only stands out when the big Polylith logo is present.

doc/shell.adoc
lread: Thanks, I was thinking these particular examples don't really need to be screenshots.
None of them include the big Polylith logo. We do miss out on the cursor, but at this point in the docs, that is probably not terribly important?

I suppose we could use the unicode █ if we really want that cursor in there.

```
example$ hello █
```
What do you think?

tengstrand: I think they are fine as they are, and we can skip the extra cursor also.

doc/shell.adoc
lread: Added a todo to #348

doc/shell.adoc
lread: Oh but we are talking about usermanager for this example, so I'll use `usermanager.edn`.

And this affects a shell screenshot, I'll just use text for it instead.

doc/shell.adoc
lread: Cool!

lread:
@tengstrand I responded to all review feedback addressed to me!

tengstrand (approved):
Looks great!

PR #359, opened 2023-10-06
docs: fix cljdoc versions link

lread:
Should be local preview and cljdoc.org friendly now.

Todo from #348


lread:
Oops, against wrong branch.

PR #364, opened 2023-10-08
introduction doc: tutorial concept

lread:
Add concept that code examples provide an ongoing tutorial.

This should now provide something we can link back to from other docs that include tutorial code examples.

Tried to make minimal changes as I know Joakim wants to maintain a specific tone to the introduction.

Fix link to end result of tutorial so that it links back to GitHub released revision.


tengstrand (approved):
Looks good!

PR #365, opened 2023-10-08
Remaining doc todos sweep

### LINKED ISSUE #348: doc: minor update reminders

lread:
After review and merge, this should close #348.

Please take a look at individual commits for details.


tengstrand (approved):
Amazing work! I will continue by updating the code in the `issue-318e` branch to reflect that `:parameters` has changed name to `:arglists`.

PR #368, opened 2023-10-11
Review and Update Tools.deps doc

lread:
The clojure vs clj discussion now also covers running clojure as a tool.

Now linking to install docs instead of reiterating what was said therein.

Moved startup time note to install docs.

Also: some end-of-line whitespace trimming (I think I accidentally added this on my first edits).

### DIFF
--- a/doc/install.adoc
+++ b/doc/install.adoc
@@ -13,24 +13,33 @@ h| Example usage:
 | `poly help`
 | `clojure -M:poly help`
 a| `clojure -Tpoly help` +
-⚠️ xref:clojure-cli-tool.adoc[command line uses Clojure Tools exec args syntax] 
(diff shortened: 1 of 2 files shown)


doc/install.adoc
```diff
@@ -13,25 +13,34 @@ h| Example usage:
...
+| Any directory
 
 h| Switch `poly` version via:
-| `@<version>` if using HomeBrew 
-| Updating dependency for `:poly` alias in `./deps.edn` 
+| `@<version>` if using HomeBrew
+| Updating dependency for `:poly` alias in `./deps.edn`
 | Reinstalling with Clojure Tools
 
+h| Pre-compiled for faster startup? link:#startup-time[^1^]
```
tengstrand, line 33: This new row/section is great!

tengstrand (approved):

PR #369, opened 2023-10-11
Review and update Build doc

lread:
Mention tutorial.

Mention what Clojure tools.build is.

Add call-outs to build.clj.

Remove `io.github.seancorfield/build-uber-log4j2-handler` from docs and example build.clj. It is a useful lib but not specifically needed for our example.

Point to `build.clj` associated with release instead of master HEAD version.

Don't assume everybody is using Cursive: move help to tip.

Edits for clarity.

### DIFF
--- a/doc/build.adoc
+++ b/doc/build.adoc
@@ -1,70 +1,97 @@
 = Build
 
-The `poly` tool doesn’t include a `build` command.
-That’s because we don’t want the tool to restrict our build pipeline in any way.
-Instead, the tool lets us choose our own way to build our Polylith artifacts for our particular pipeline; which could be with simple build scripts, all the way to cloud-based build tools.
+The `poly` tool doesn't include any `build` commands.
+That's because we don't want `poly` to restrict your build pipeline in any way.
+Instead, `poly` lets you choose how to build your deployable artifacts for your particular pipeline.
+Whether you choose simple build scripts or cloud-based build tools, `poly` does not get in the way.
 
-Let's say we want to create an executable jar file out of the `command-line` project.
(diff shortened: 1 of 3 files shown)


examples/doc-example/build.clj
```diff
@@ -68,9 +66,6 @@
                              {:basis        (b/create-basis)
                               :class-dir    class-dir
                               :compile-opts {:direct-linking true}
-                              ;; if your project (or any of its dependencies)
```
tengstrand, line 71: Also make the same deletion from `scripts/sections/build/build.clj`.

tengstrand (changes requested):

doc/build.adoc
```diff
@@ -1,70 +1,97 @@
 = Build
 
-The `poly` tool doesn’t include a `build` command.
-That’s because we don’t want the tool to restrict our build pipeline in any way.
-Instead, the tool lets us choose our own way to build our Polylith artifacts for our particular pipeline; which could be with simple build scripts, all the way to cloud-based build tools.
+The `poly` tool doesn't include any `build` commands.
+That's because we don't want `poly` to restrict your build pipeline in any way.
+Instead, `poly` lets you choose how to build your deployable artifacts for your particular pipeline.
+Whether you choose simple build scripts or cloud-based build tools, `poly` does not get in the way.
```
tengstrand: Maybe use doesn't.

lread: Will do.

examples/doc-example/build.clj
lread: Ah did not see that one, will do.

lread:
@tengstrand: all review feedback addressed!

tengstrand (approved):

PR #370, opened 2023-10-13
Review and update Git and Tagging docs

lread:
The Git doc seemed to focus on italicized "stable point in time". It seemed important, so I carried the term through to the Tagging doc.

While walking through Git doc examples, it became obvious to me that I had missed the tutorial step of creating lisa.clj from the Development doc. This was because the step was buried in Cursive setup instructions. I've teased out a Generic Setup section for non Cursive users.

Changed git log cmd in some cases for a less wide and more doc friendly output.

And: a bunch of edits for clarity.

### DIFF
--- a/doc/workspace.adoc
+++ b/doc/workspace.adoc
@@ -135,6 +135,7 @@ my-git-repo-dir
 └── workspace.edn
 ----
 
+[[workspace-root-under-git-root]]
 ...or you create the workspace in a directory under the git repository root by executing e.g.,:
 
 [source,shell]
--- a/doc/development.adoc
+++ b/doc/development.adoc
@@ -1,4 +1,5 @@
 = Development
+:toc:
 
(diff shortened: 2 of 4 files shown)


doc/workspace.adoc
```diff
@@ -135,6 +135,7 @@ my-git-repo-dir
 └── workspace.edn
 ----
 
+[[workspace-root-under-git-root]]
 ...or you create the workspace in a directory under the git repository root by executing e.g.,:
```
tengstrand, line 140: In the "NOTE: In the above examples you'll notice we ommitted the `:commit` option." below, we could mention that it's possible to give `:add` instead of `:commit` which will at least add the newly created files to git, so they only have to commit them (not add + commit). This is not super important, but was something that I realized.

tengstrand (approved):
Looks great!

PR #372, opened 2023-10-17
Add Greenlabs to production-systems doc

namenu:
Add a new case to the Production Systems page.

Let me know if there's anything else you'd like me to include.

cc. @yyna


doc/production-systems.adoc
```diff
@@ -49,6 +49,15 @@ We made our initial commit on June 9, 2019.
 The Polylith way of working gives us a range of benefits, such as decoupled building blocks that are easy to find, change, reuse, and reason about.
 We combine these building blocks to create the `poly` and `polyx` command line tools, and the `clj-poly` library.
 
+== Greenlabs
+
```
tengstrand: It would be good if you could include the image also (see how it's done by the other companies), e.g.:

image::images/production-systems/greenlabs.png[link="images/production-systems/greenlabs.png"]

tengstrand: It would also be good if you could move it up, so that it comes after World Singles Networks, but before Polylith.

It would also be nice if you could add aliases to your projects (but that's up to you) to make the projects more readable, e.g.:
- `aws-lambda-braze` could have the alias `brz` or `braze`
- `aws-lambda-cach-receipt-update` could get the alias `cru`
- and so on...instead of showing the default values `?1`, `?2`, `?3`...

tengstrand (changes requested):

doc/production-systems.adoc
namenu: Normally, I removed aliases because they don't seem to improve readability, but in this case I agree that the image visibility wasn't good. Added aliases along with gpt :)

tengstrand: Great! You use different version of the same libraries, e.g. buddy-sign, com.cognitect.aws/api, and others. I have no problems with leaving it that way, but it would look better if you use the same version of each library everywhere.

namenu: Yeah, diverging library versions is the problem we have..
I can make it more consistent, but it think it will take some time.

tengstrand: I merge this, and then you can add another PR later! It's an impressive workspace you have. Really cool that you wanted to share it.

PR #377, opened 2023-10-29
Review & Update Flags, Testing, Test Runners docs

lread:
General
- Noticed I had accidentally changed `command-line` project alias from `cl` to `cli`, so addressed that in project.adoc.
- Edits for clarity.
- Info snapshot images are now often annotated, usually just with an orange box to focus the reader on the part of the output we are describing. Since the snapshots images are generated automatically via a script, I've decided to link to the generated image from the manually annotated InkScape svg. I'll circle back in a future PR to look at the script that generates these snapshots and see what we can improve.

Flags doc
- Describe flags in general
- Add section headings and toc
- Re-order brick flags content to match order of project flags
- Do more hand-holding in flags with the help of new `info` output images that point to what we are describing.

Testing doc
- Added some upfront description on how poly runs tests which should help newcomers.
- Annotated info images with focus boxes
- Fixed a fn name typo in setup/teardown examples

Test Runners doc
- re-orged to cover what the average reader might be first interested in
- used Sean's and Imre's test runners in examples so reader can actually try something concrete out.
- clarified that poly only includes an in-process test runner


lread:
Ok @tengstrand, this is a big one, and I expect it will take a while for you to review.
Please review the commit comment for an overview.
I filled in information gaps with my understanding; please review carefully to make sure my understanding is actually correct!

doc/flags.adoc
```diff
@@ -1,80 +1,180 @@
...
 
-[source,clojure]
-----
- :aliases  {:dev {:extra-deps {poly/user {:local/root "components/user"}
-  ...
-            :test {:extra-paths ["components/user/test"
-----
+The `dev` flags `---` for `command-line` means `poly` sees:
+
+. `./projects/command-line/src` isn't referenced in the workspace `./deps.edn`
```
tengstrand: Should be: isn't referenced in the development `./deps.edn` file (or similar, not workspace).

doc/flags.adoc
```diff
@@ -1,80 +1,180 @@
...
+. `./projects/command-line/test` isn't referenced in the workspace `./deps.edn`
```
tengstrand: development, not workspace

doc/flags.adoc
```diff
@@ -1,80 +1,180 @@
...
+:toc:
 
-We have one more thing to cover regarding the xref:commands.adoc#info[info] command, and that is what the `-`, `s`, `t` and `x` flags mean:
+You've seen output from the xref:commands.adoc#info[info] command, but we haven't explained flags:
 
-image::images/flags/projects.png[width=350]
+image::images/flags/info-flags.png[width=475]
 
-Let's start with the _status_  column:
+There are four possible flags:
```
tengstrand, line 8: For me this became more complicated, that we now try to introduce both the project flags and the brick flags at the same time. I like the boxes around each area, but maybe we could introduce projects first (with just one box) and then the bricks section later, also with one box.

If we do that, we could be more precise in the descriptions.
Then we could say things like:
`s` means a project's `src` directory is included in the project, otherwise `-`.
`t` means a project's `test` directory is included in the project, otherwise `-`.
 
In the bricks section we can say:
`s` means a component's `src` directory is included in the project, otherwise `-`.
`t` means a component's `test` directory is included in the project, otherwise `-`.

When we create a brick, the tool creates a `src`, `resources` and `test` directory for us, and a `deps.edn` file, where the `src` and `resources` directories are specified in `:paths` as `["src" "resources"]`, and the `test` directory as `["test"]` in `:aliases > :test > :extra-paths`. 

But we are free to change the name of the `src` directory, or store the source in more than one directory, e.g.
`["src2" "cljc" "resources"]`, or `["test1" "test2]` for the test context.

The `resources` directory is an exception, which has to have exactly that name.

I suggest that we talk about `src` and `test` everywhere, to make it simpler, and then at the end we can have an info box that clarifies that it's more flexible.

Also, talking about `true` and `false` maybe just make the reader more confused. Maybe we should change from "Flags" to "Status flags" or something, but I'm not sure about that.

tengstrand (changes requested):

doc/flags.adoc
lread: Because there are many `deps.edn` files, my feeling is that using `./deps.edn` isn't clear enough for the reader. I've been calling this `deps.edn` the "workspace `./deps.edn`", [example](https://github.com/polyfy/polylith/blob/b3dcc8bbb315dfd2a9b8d011c3195c4b617ffd0b/doc/project.adoc?plain=1#L8). Are you suggesting we call it the "development `./deps.edn`"? Or maybe "development project `./deps.edn`"?

doc/flags.adoc
lread: see above reply

doc/flags.adoc
lread: > For me this became more complicated, that we now try to introduce both the project flags and the brick flags at the same time. I like the boxes around each area, but maybe we could introduce projects first (with just one box) and then the bricks section later, also with one box.

And

> Also, talking about true and false maybe just make the reader more confused. Maybe we should change from "Flags" to "Status flags" or something, but I'm not sure about that.

Hmm... I'm not sure I understand the confusion yet.

We can assume a `poly` user is a programer and clearly understands booleans and `true` and `false`. Do you think maybe because you've been working on the higher level docs there is a chance you might be thinking from a different perspective?

Coming in as a newbie, I have absolutely no clue what flags are. 
Do you think it is confusing to describe them as a general concept first and then ease into the details?

> When we create a brick, the tool creates a src, resources and test directory for us, and a deps.edn file, where the src and resources directories are specified in :paths as ["src" "resources"], and the test directory as ["test"] in :aliases > :test > :extra-paths.

> But we are free to change the name of the src directory, or store the source in more than one directory, e.g.
["src2" "cljc" "resources"], or ["test1" "test2] for the test context.

> The resources directory is an exception, which has to have exactly that name.

> I suggest that we talk about src and test everywhere, to make it simpler, and then at the end we can have an info box that clarifies that it's more flexible.

Yes, good points. I notice that somewhere else the docs talk about the src and test "contexts". But I was left guessing what these terms meant. Are there terms the Clojure core team uses for this concept? 
I guess technically, it comes down to the classpath, yes? The default classpath and the `test` alias classpath?

doc/flags.adoc
tengstrand: Yes, "development project `./deps.edn`" sounds good to me.

doc/testing.adoc
```diff
@@ -243,24 +321,26 @@ example
...
 [source,clojure]
 ----
- :aliases {:test {:extra-paths ["test"]
+ :aliases {:test {:extra-paths ["test"] ;; <1>
                   :extra-deps  {}}
 ----
+<1> Add `test` path
 
-...and to `./deps.edn`:
+Now add this same path to your workspace `./deps.edn`:
```
tengstrand: Change to "development project `./deps.edn`

doc/flags.adoc
lread: Sounds good, will update.

lread: Hmmm.... as I am making this change, I'm finding it might confuse the reader who is not very familiar with what the development project is. They might assume it is in some other dir. 

Although incorrect, "workspace `./deps.edn`" implied its location.

I think your original strategy of just `./deps.edn` (without any descriptors) might be the least confusing. Unless you have other ideas?

tengstrand: It is the config file for the development project. Maybe people know that already, so then we may skip that information.

doc/flags.adoc
lread: @tengstrand So, I think after Slack discussions, and input from others, we are now OK with the general description of flags?  

I think your feedback about the `src` and `test` paths is still worth exploring.
I can try to work on a change that explains this.

tengstrand: Yes, let's go with the general description of flags. 
"I can try to work on a change that explains this." - yes if you want to. I still feel that we can be more crisp about this.

doc/flags.adoc
lread: Ok, from Slack, @tengstrand has agreed to go with an unadorned `./deps.edn`.

doc/flags.adoc
lread: Added note on src and test paths.

lread:
Ok @tengstrand, all known/remembered feedback addressed.

doc/flags.adoc
```diff
@@ -1,80 +1,183 @@
...
 
-[source,clojure]
-----
- :aliases  {:dev {:extra-deps {poly/user {:local/root "components/user"}
-  ...
-            :test {:extra-paths ["components/user/test"
-----
+The `dev` flags `---` for `command-line` means `poly` sees:
+
+. `./projects/command-line/src` isn't referenced in ./deps.edn`
```
tengstrand: Missing back-tick for ./deps.edn

tengstrand (changes requested):

doc/flags.adoc
lread: Tx

doc/testing.adoc
lread: Pretty sure this is done with prev commit.

tengstrand (approved):
Thanks for this!

PR #385, opened 2023-11-10
Rewrite create-*.sh bash scripts in babashka

lread:
Convert create-example.sh, create-example-images.sh, help.sh, help-doc.sh, update-commands-doc.sh to bb tasks. (Also deleted stale local-dep-old-format.sh and local.sh which had been incorporated into create-example.sh)

Scripts now also run on Linux.
The bash scripts used `sed` and `mktemp` in ways that only worked on macOS. Since we are now using babashka, we no longer need to make use of these cmds and are therefore now support both macOS and Linux.

New bb tasks:
- clean - deletes build work (including tmp dir work)
- poly-uberjar - creates poly jar, but only if sources have updated (replaces old jar task)
- polyx-uberjar - create polyx jar, but only if sources have updated
- help-txt - generates doc text files via poly help
- help-doc - assembles doc text files into doc/commands.adoc
- update-commands-doc - replaces scripts/help/update-commands-doc.sh (runs help-txt then help-doc)
- create-example - merges scripts/create-example.sh and scripts/create-example-images.sh (does not call update-commands-doc)
- gen-all - replaces scripts/create-example.sh, scripts/create-example-images.sh (depends on update-commands-doc)

Nitty Gritty fixes:

components-tree.txt now generating to output dir (was generating to temp dir)

no longer generating extra overview.png and info.png to scripts/output dir (still generating to images dir)

add missing txt output for (only image was being generated): poly info :no-changes brick:cli :all-bricks fake-sha:e7ebe68

add `:no-changes` to polyx overview.png to match Polyx docs.

clearly distinguish generated images from hand-crafted ones: add `output` dir before png filename.

Other changes:

For each spawned command now clearly logging cwd, command, and stdout redirection (if specified).

I felt automatically updating clojure to the latest version was beyond the scope of what this script should do. It also assumed brew usage and therefore macOS. We now instead fail fast if the installed version of clojure is not the latest.

Now explicitly checking for prerequisite tools: `clojure`, `git`, `tree`. We assert for a min version of `tree`, older versions counted dirs differently.

Now generating a single root tmp work dir for all tmp generated work. This makes review and cleanup easier.

Now building poly and polyx jars and invoking them via `java -jar` in the same way a stand-alone install would and avoids manual pre-setup required by old bash scripts.

Because the poly and polyx jars take a while to build, only rebuilding if there have been changes.

Dir tree output now includes distinguishing trailing forward slash for directories.

Change test output to match new formatting in docs:
- was: `exit code: 0`, now: `# Exit code: 0`
- output now always gets same treatment (sometimes Execution time was not normalized in bash scripts)

Clone from GitHub via https protocol. This allows the script to be run by folks who have not setup their git ssh keys.

Modified gen-nav to print filename it is generating to stdout. This removes unnecessary mystery for those unfamiliar with what gen-nav does.

Added --no-pager option with using `git log`.
I was finding pager was being invoked on during some macOS testing.


lread:
@tengstrand here's what I've got so far.
Please take time to read the commit message, it should answer many questions.

bb.edn
```diff
@@ -17,12 +18,40 @@
...
+  clean {:doc "Delete build work and artifacts"
+         :task clean/-main}
+
+  doc-preview {:doc "Preview what docs will look like on cljdoc, use help for args"
                :task cljdoc-preview/-main}
 
-  jar {:doc "Builds projects/poly/target/poly.jar"
-       :task (shell "clojure -T:build uberjar :project poly")}
+  poly-uberjar {:doc "Builds poly uberjar, if needed (run clean first to force)"
+            :requires [poly-jar]
```
tengstrand, line 28: The formatting looks strange here, and on row 32 and 33..

scripts/shell.clj
```diff
@@ -0,0 +1,62 @@
...
+  - pretty print out each cmd before it is run
+  - support altering output via :alter-out-fn (assumes :out is file)"
+  [& args]
+  (let [[opts & args] (if (map? (first args))
+                        args
+                        (conj args {}))
+        {:keys [out dir alter-out-fn]} opts
+        opts (assoc opts :pre-start-fn (fn [o]
+                                         (let [cmd-str (->> (:cmd o)
+                                                               (mapv #(if (str/includes? % " ")
```
tengstrand, line 24: Wrong indentation, `(mapv` should be aligned with `(:cmd`.

tengstrand (changes requested):
This is amazing! Good job!

bb.edn
lread: ah yes, tx, I'll run cljfmt on the files I changed, there are a number of formatting issues!

tengstrand (approved):

PR #389, opened 2023-11-15
Review and update Dependencies doc

lread:
Add more sections and toc.

Moved notes on ws-dir to Exploring the Workspace doc (as adoc comment for now, I'll integrate when I get to that doc).

Now generating screenshots with `polyx` from `create_example.clj`.

In Clojure we typically think of "dependencies" as 3rd-party library dependencies. Because "dependencies" means something different in the world of poly, I added more words here and there distinguishing the difference.

Annotated first deps screenshot to emphasize uses vs used-by.

Doc talked about what yellow means but did not mention other colors, so I filled in that gap.

Switch from ssh to https for git clone so that clone will work for readrs that don't have ssh creds setup.

Other edits for clarity.


lread:
@tengstrand Ready for your review and feedback.
As usual, please read over the commit message first for an overview of the changes I made.
Also, please give a little extra time to the Accessibility section to make sure the changes I made to the Caution column are correct.

doc/dependencies.adoc
```diff
@@ -13,129 +18,131 @@ clone-from-here
...
 
 [source,shell]
 ----
-cd ../..
-poly info
+poly deps
 ----
 
-image::images/dependencies/info.png[width=400]
+image::images/dependencies/deps-interfaces-annotated.png[width=650]
```
tengstrand, line 48: Maybe it's enough to say "article component row" and "article interface column" in the picture (in orange text), but skip the rest of the text. I should be enough I think to only have it in the text below the picture.

tengstrand (changes requested):

doc/dependencies.adoc
lread: Here's the image for reference:
image

lread: I think I disagree on this one.  The "uses" vs "used by" and how they map to rows vs columns is an important (and not obvious) concept to convey. I'm also OK with repeating diagram text in plaintext because some folks can't see and use screen readers.

Can you tell me more about why you think the text should be changed?

tengstrand: I wasn't sure about this, and I think it's fine to keep (almost) the same text in two places!

PR #391, opened 2023-11-18
Review and update Libraries doc

lread:
Generate libs, libs :compact, and libs with overriden dep from create_example.clj.

Add support to create-example to run a subset of tasks by a group id. Running all tasks can take several minutes, so it is handy to be able to, for example, run `bb create-example realworld` to only run the realword tasks. If no group ids are specified, all tasks are run.

Added more sections and a table of contents.

The word "dependencies" is overloaded because we call a brick a "dependency" but a 3rd-party library is also a "dependency". I tried to soften the confusion by adding a few words.

Re-organized content to start with something the reader can easily try out then dug progressively into details.

Edits for clarity.


lread:
@tengstrand Here's my crack at the Libraries doc.

doc/libraries.adoc
```diff
@@ -56,53 +91,116 @@ We can ignore updating libraries for a brick or project by adding the `:keep-lib
...
+If you want to run your tests directly from a project using the tools.deps CLI tool, you must duplicate test library dependencies from the bricks to the project's `deps.edn` file under `:aliases > :test > :extra-deps`.
+If you only run tests with the built-in xref:commands.adoc#test[test] command, you don't have to worry about this.
+
+=== Project Library Dependencies
+
+The library dependencies for a project are the sum of all library dependencies that are indirectly included via its bricks, together with dependencies declared by the project itself.
+If different versions of the same library dependency exist, then `poly` will use the latest version.
+
+You override the library version used for a project in its `deps.edn` file.
+For example:
```
tengstrand, line 199: Looks like there is a missing example here.

tengstrand (changes requested):
Except from one thing, everything looks great!

doc/libraries.adoc
lread: Oh thanks! I accidentally left the old example prompt in there with my simplified replacement. Will fix.

lread: Fixed!

lread:
@tengstrand I've addressed your review feedback.

doc/libraries.adoc
tengstrand: Cool!

tengstrand (approved):

PR #398, opened 2023-11-22
Review and Update Source Code doc

lread:
Added table of contents.

The first block of text talks specifically about multiple paths but had no heading, so added one.

I downplayed .clj vs .cljc code organization in first section to be an illustrative of multiple paths. (To me, the text seemed to be recommending separate `clj` and `cljc` dirs was some sort of poly recommendation, which felt odd.)

Reworded a some text under the frontend section.
The term "stripped" did not mean much to me, so I tried to instead explain.
The term "import" was problematic because it will be confused with Clojure's `import` so reworded that too.

Other edits for clarity.

### DIFF
--- a/doc/source-code.adoc
+++ b/doc/source-code.adoc
@@ -1,12 +1,17 @@
 = Source code
+:toc:
 
-In all our examples so far, the source code has been stored in two directories, `src` and `test`.
-xref:tools-deps.adoc[Tools.deps] supports splitting up the source and test code into several directories, and that is also supported by the `poly` tool.
+== Paths
 
-It's also worth mentioning that the `poly` tool understands `.cljc` files which can be used to share code between `.clj` and `.cljs`.
+In all our examples thus far, we've shown source code stored in two directories: `src` and `test`.
+xref:tools-deps.adoc[Tools.deps] supports splitting source and test code into several directories, as does `poly`.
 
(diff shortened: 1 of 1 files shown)


lread:
@tengstrand, ready for your review and feedback.

doc/source-code.adoc
```diff
@@ -59,64 +66,74 @@ For components, the name is the component's xref:interface.adoc[interface] name,
...
-The `poly` tool can handle both situations, so it's more a matter of preference what you choose.
+The last alternative is to put your frontend code in a separate git repository.
 
-==== 3. Let the frontend code live in another repository
+Some downsides to this strategy are:
 
-The last alternative is to put the frontend code in a separate repository.
-If we do that we will lose the Polylith way of sharing "living" code, by automatically having access to the latest version of the code.
-Now we have to "freeze" the `.cljc` code by building a library of it, to be able to share the code.
+* You are no longer working with a monolith and might even have separate frontend and backend releases
```
tengstrand: I suggest we change from `monolith` to `monorepo`.

tengstrand (changes requested):

doc/source-code.adoc
lread: Yeah, good idea, will do.

lread: Done!

PR #399, opened 2023-11-22
Review and update Artifacts doc

lread:
Add a section for Libraries and a sub-section for caveats when publishing multiple libraries.

Turf the paragraph on the advantages of multiple libraries. I felt it was a stretch and wishful thinking, something that readers would see through and won't appreciate.

Add more links to other docs.

Edits for clarity.


lread:

tengstrand (approved):

PR #404, opened 2023-11-24
Review and update Validations doc

lread:
Re-organize and re-section.

Add some info on exit codes.

Add a screenshot to show an example of `ws get:messages`.

Update Warning 206 screenshot to what it should be now: Error 111.

Turf note on Warning 207 that it is only be applied to check command, it's not true anymore.

Configuration doc:
- the only reason we documented `:bricks` was to describe disabling Error 111, this is no longer a feature so we no longer need to describe `:bricks`.
- correct  description for `:necessary` and add link to Validations doc.

Edits for clarity.


lread:
As always, please read the commit message first.

doc/configuration.adoc
```diff
@@ -89,10 +89,6 @@ The workspace map keys are:
 |===
 | Key | Description
 
-a| `:bricks`
```
tengstrand, line 92: We need to keep this one, and add information about the `:keep-lib-versions` key, which is described in libraries.adoc. Probably a link to the library section is enough.

doc/configuration.adoc
```diff
@@ -168,7 +164,8 @@ You can use the alias instead of the full project name for various `poly` comman
 If not specified, the xref:commands.adoc#info[info] and xref:commands.adoc[libs] commands emit a `?` followed by a number (e.g. `?1`) for project column headings.
```
tengstrand, line 172: We need to add `:keep-lib-versions` here (that I missed). We could add the key to the example above also.

doc/configuration.adoc
```diff
@@ -168,7 +164,8 @@ You can use the alias instead of the full project name for various `poly` comman
...
 
 a| [nowrap]`:necessary`
-a| Supports turning off `Error 111` for one or more bricks.
+a| Supports turning off `Warning 207` for one or more bricks.
```
tengstrand, line 177: Good catch!

tengstrand (changes requested):

doc/configuration.adoc
lread: Aha, ok, will do.

doc/configuration.adoc
lread: Righty-o, will do.

doc/configuration.adoc
lread: tx

doc/configuration.adoc
lread: Done!

doc/configuration.adoc
lread: Done!

lread:
@tengstrand review feedback addressed.

tengstrand (approved):

PR #405, opened 2023-11-24
Turf docs on Clojure Tools CLI Usage

lread:
Joakim surveyed the community and they responded that they are good with removing using `poly` as a Clojure Tool.

Seems like a good idea, the mismatch between argument parsing was awkward and just one more thing to learn.

This change reflects docs only and does not remove supporting code.

Also: noticed and turfed an old println used for debugging in create_example.clj


tengstrand (approved):
Looks good!

PR #407, opened 2023-11-25
Review and Update Naming doc

lread:
Add a quote on naming.

Add links to other docs.

Naming is subjective. Change tone from "this is a good way to name" to "consider naming like so".

Reorder advice on multiple interfaces with same name to come after more generic advice on components.

"For tools, we can use the tool name" did not give any naming advice at all, so replaced it with a suggestion.

Edits for clarity.

### DIFF
--- a/doc/naming.adoc
+++ b/doc/naming.adoc
@@ -1,40 +1,49 @@
 = Naming
 
-Every time we create an `interface`, `component`, `base`, `project` or `workspace`, we need to come up with a good name.
-Finding good names is one of the hardest and most important thing in software.
-Every time we fail to find a good name, it will make the system harder to reason about and change.
 
-The components are the core of Polylith, so let's start with them.
-If a component does _one thing_ then we can name it based on that, e.g. `validator`, `invoicer` or `purchaser`.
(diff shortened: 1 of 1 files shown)


lread:

doc/naming.adoc
```diff
@@ -1,40 +1,49 @@
...
 
-Every time we create an `interface`, `component`, `base`, `project` or `workspace`, we need to come up with a good name.
-Finding good names is one of the hardest and most important thing in software.
-Every time we fail to find a good name, it will make the system harder to reason about and change.
 
-The components are the core of Polylith, so let's start with them.
-If a component does _one thing_ then we can name it based on that, e.g. `validator`, `invoicer` or `purchaser`.
-Sometimes a component operates around a concept that we can name it after, e.g. `account` or `car`.
-This can be an alternative if the component does more than one thing, but always around that single concept.
+"There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors."
```
tengstrand, line 4: I like this! I have had this ide, to sprinkle some quotes around. I'm sure we can fit in a few more in the documentation!

tengstrand (approved):
Excellent!

PR #409, opened 2023-11-25
Review and Update Git Hook doc

lread:
Some folks find git hooks more of a nuisance than a help, so I downplayed them to make them feel like an option rather than a recommendation.

Example hook now uses the `:poly` alias for a simpler script that can be copy pasted.

Mention example hook script is OS-specific.

Add some warnings and tips.

Link to other docs.

Edits for clarity.

### DIFF
--- a/doc/git-hooks.adoc
+++ b/doc/git-hooks.adoc
@@ -1,18 +1,27 @@
-= Git hook
+= Git Hook
 
-We can ensure that we don't push code that puts the workspace in an invalid state, by adding a https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks[git hook] to our workspace, that executes the xref:commands.adoc#check[check] command.
+You can use https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks[git hooks] to have Git invoke custom scripts for specific Git events.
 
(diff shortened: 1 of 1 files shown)


lread:

doc/git-hooks.adoc
```diff
@@ -1,18 +1,27 @@
...
 
-Replace `PATH-TO-WORKSPACE-DIRECTORY` with the path to the workspace root.
+TIP: Don't forget to `chmod +x ./get/hooks/commit-msg`.
+
+TIP: Client-side git hooks are developer-specific and not tracked by `git`.
+
+WARNING: Failing hooks can be mysterious when committing from an IDE.
+If the commit hook fails, the commit will fail, but you might not get a meaningful message as to why.
+
+TIP: Consider invoking a xref:install.adoc[poly stand-alone AOT-compiled jar] for a faster `check`.
```
tengstrand: This page is about things we do when committing files, so I'm not sure if this tip should be here. If we want to keep this tip, then we should also tell the user that another option is to start a shell.

tengstrand (changes requested):

doc/git-hooks.adoc
lread: @tengstrand, I clarified that the tip is talking about the git hook script. 
Good for you now?

tengstrand (approved):

PR #410, opened 2023-11-26
Turf Context doc

lread:
I shared with Joakim that I found this doc confusing because it does not talk about polylith at all. It is geared toward object oriented developers coming to Clojure and what that means in an IDE.

He asked the polylith community what they thought and they agreed the doc could be deleted.

Note: when deleting `context` from `doc :page` output example, I went from 2 cols to 3. I thought this was a better use of vertical space.


lread:
@tengstrand, turfing ready for review!

tengstrand (approved):

PR #411, opened 2023-11-27
Review and update Colors doc

lread:
Add toc and a section.

Show colors first.
Give examples for what colors are used for.
Include ANSI color codes (and names where they differ) to give readers more of chance to be able to configure for their terminal. Explain that sample colors represent a terminal config that the poly developers like (before these seemed to be showing the absolute colors poly would emit, which is not the case).

Verify sample RGB colors and regenerate images with ImageMagick in `polylith.clj.core.util.interface.color` via temporary code:

```clojure
(comment
  (require '[clojure.java.shell :as shell])

  (run! (fn [[color [r g b]]]
          (let [hex-color (format "#%02x%02x%02x" r g b)]
            (println color hex-color)
            (shell/sh "convert" "-size" "38x38" (format "canvas:%s" hex-color)
                      (format "doc/images/colors/%s.png" (name color)))))
        colors)

  nil)
```

Correct filename tyo: yelow.png -> yellow.png

No longer mislead about the power of `dark` and `light`. They have very little effect and are minor tweaks for terminal color theme configuration.

Polyx: mention that images always use a dark-themed output.

Edits for clarity.


lread:

tengstrand (approved):
This was an easy one! ;-)

PR #413, opened 2023-12-01
Review and update Continuous Integration doc

lread:
After me discussing with Joakim and then he with Sean, we found that tracking per project did not make sense. So it that content has been turfed.

Described 2 simple tagging/release strategies for readers to consider.


lread:
@tengstrand, ready for your review and feedback.

tengstrand (approved):
Looks great!

PR #426, opened 2024-01-26
link to affected 3rd party test runners in docstring

seancorfield:
Signed-off-by: Sean Corfield <sean@corfield.org>


imrekoszo:
@seancorfield thanks for documenting this.

Do you think it would be worth making the contract explicit? If yes, have you got any suggestions on a way to do that?

seancorfield:
Specs for the arguments that cover the required keys and an `s/fdef` on `create` itself -- with instrumentation enabled for the tests (I'm assuming there are some tests for `create-test-runner`).

For now, just a great big reminder in the docstring is probably enough :)

tengstrand (approved):
Looks good!

PR #482, opened 2024-07-04
Create an example project showing how to use Polylith with an Integrant system

marksto:
Hi Joakim @tengstrand!

Here you are, an example project that we discussed a few days ago in Slack.

The example rests in the `examples/integrant-system` directory. I've added both tests (in `bases/system/test/integrant/system/core_test.clj`) and docs (in the `readme.md`) for it. I also decided to keep it simple and didn't introduce another `project`, since it would be a complete copy of the `development` one.

Please, feel free to propose and/or introduce any changes.

PR #486, opened 2024-07-07
Clarify the example project showing how to use Polylith with an Integrant system

marksto:
Hi Joakim!

Here's a clarification of Whats and Whys for the **Polylith + Integrant** system example introduced in #482. I decided to write them down after having a conversation with folks on Slack.

Also, it now caters for in-REPL development with `integrant/repl` lib that I forgot to add in the first PR (even though I initially intended to do so and introduced a dependency on the lib previously).

Cheers,
Mark


tengstrand:
This is really valuable. Thank you so much @marksto!

PR #501, opened 2024-07-23
Add diagrams, generated by git-of-theseus

tengstrand:
Add git-of-theseus section in the "Developing poly" page.


tengstrand:
The poly tool now includes .cljs code, so we need to work a bit more on this before it's time to release even a snapshot version.

Issue #516, opened 2024-11-29
More control over created brick contents


seancorfield:
**Is your feature request related to a problem? Please describe.**
I would like to be able to control the contents of newly-created bricks.

Specifically:
* a way to specify/override the default version of Clojure used (it's currently hard-coded)
* a way to suppress creation of `resources` (I very rarely need this, and I don't like it being created by default -- and the `.keep` file should probably be in `resources/<brick-name>/` instead of just `resources`)
* a way to provide a standard prelude in new `src`/`test` files (e.g., a copyright message)
* a way to specify different requires in `test` files and a different default testing library

**Describe the solution you'd like**
Ideally, use [Selmer](https://github.com/yogthos/Selmer) templates for various files that need to be created, with a way for users to provide their own on a global or per-workspace basis.

Provide a way to configure (in `workspace.edn` and via the CLI) additional/overriding "variables" to be substituted into these templates.

Perhaps, when a workspace is created, Polylith could copy its default templates (or a user's global overrides?) into a folder, and those templates would be used for any brick creation from then on, simplifying any customization process?

seancorfield:
Even if this is some ways off, a couple of changes should be made sooner:

* update the `shared/clojure-ver` from `1.11.1` to `1.12.0` (note: 1.11.4 is the recommended version of 1.11 anyway)
* add an additional `.gitignore` for Calva's new REPL window: `.calva/repl.calva-repl` (`.calva/output-window/output.calva-repl` should be kept as the old ignore too, I think)

I've submitted PRs for both of these.

seancorfield:
Weird, I thought I'd written a complete proposal of the template structure but I can't find it, so I'll add it again.

I'm proposing a structure like the following:

```
templates/
  .gitignore
  deps.edn
  readme.md
  workspace.edn
  bases/
    deps.edn
    main.clj
    test.clj
  components/
    deps.edn
    interface.clj
    impl.clj
    test.clj
  projects/
    deps.edn
```

If `impl.clj` is present, it is used as a template for `components/<entity>/src/<top-ns>/<entity>/impl.clj` so users can get started with a split interface/implementation component if they want.

The defaults would be in Polylith's `creator` brick under `resources/creator/templates` and users could have "global" overrides in `{user.home}/.config/polylith/templates` (alongside the user `config.edn`). Per-project overrides could be under `.polylith/templates` in the workspace.

seancorfield:
[...]

PR #520, opened 2024-11-30
Using Selmer templates for create fragments

seancorfield:
This is a first cut PR for discussion and feedback.

It addresses #516 by converting all of the hardcoded file contents in the `creator` brick to Selmer templates.

What is not included:

* tests for overriding the defaults at the user or project level
* setting to suppress `resources` generation (it's less important and it's orthogonal to the template work anyway)
* any attempt at documentation -- there doesn't seem any point until we've agreed on the mechanics of this


seancorfield:
I've updated our work project to use this PR and added project-level templates and tested that level of override works well.

tengstrand (approved):

Issue #448, opened 2024-03-08
Support scanning tests in src directories


tengstrand:
People sometimes put test code together with the src code. Today we don't run those tests. It would be good if this could be configurable.

To turn it on globally, we could put an `:include-src-dirs` flag in `workspace.edn` at the root to affect all tests:

```
{
 ...
 :test {:include-src-dirs true}
}
```

or per project:
```
{
 ...
 :projects {"myproject" {...
                         :test {:include-src-dirs true}}
}
```

or even per brick:
```
{
 ...
 :bricks {"mybrick" {...
                     :test {:include-src-dirs true}}
}
```

marksto:
Totally support this feature of `poly test`, but I must admit that I never run into [the problem](https://github.com/polyfy/polylith/blob/master/examples/poly-rcf/readme.md) myself, neither with [rcf](https://github.com/hyperfiddle/rcf) nor with [rich-comment-tests](https://github.com/matthewdowney/rich-comment-tests) on a Polylith-based project that used both.

seancorfield:
I use rich-comment-tests (per marksto above) but we follow the recommendation in RCT to have a stub file in the `test` tree that calls into the relevant source ns to actually run the inline tests as part of regular testing.

I would never want `src` folders scanned for tests but this seems like a harmless (if somewhat pointless) enhancement as long as it is **OFF** by default.

Plenty of tooling out there doesn't support tests in `src` and I call it out as a problematic approach in the Expectations documentation because of that.

tengstrand:
This issue was created to make it easier to work with [Hyperfiddle rcf](https://github.com/hyperfiddle/rcf) as described [here](https://github.com/ieugen/poly-rcf/blob/main/readme.md). As it is today, you can add the `src` directory to the `test paths` to achieve this, but making it configurable would make it easier for these users. I can't estimate how important this is and if it's worth supporting. Sean points out that it's a problematic approach. What do you say @ieugen?

ieugen:
hi @tengstrand , I am a bit unsure after reading the comments.

I will try out https://github.com/matthewdowney/rich-comment-tests and probably migrate to it - since it seems a bit less intrusive than RCF .

In gradle, src classes are available for tests on the classpath. 
Not sure why with clojure is (seems) different. 

If the feature is easy to implement and maintain than I am ok with the approach. 
As an alternative to a feature that would need to be maintained, this could be fixed via improving  documentation maybe? 

My experience with Clojure is a bit more limited than some. 
[...]

PR #527, opened 2025-01-10
Extend bricks to test calculation

imrekoszo:
Adds a new key to `project` (`bricks-to-test-all-sources`) which differs from `bricks-to-test` in that it might contain bricks that don't have any dedicated test sources.

This is to let test runners discover potential tests in src-only bricks as well. (Like Kaocha's [automatic spec test generation](https://github.com/lambdaisland/kaocha/blob/main/doc/spec_test_check.md))

TODO:

- [x] find a better name for the new key
- [x] update docs

Also see #448

### DIFF
--- a/components/test/src/polylith/clj/core/test/bricks_to_test.clj
+++ b/components/test/src/polylith/clj/core/test/bricks_to_test.clj
@@ -14,6 +14,9 @@
     (set/intersection (into #{} (mapcat :src) [base-names component-names])
                       (into #{} (mapcat :test) [base-names component-names]))))
 
+(defn bricks-referenced [{:keys [base-names component-names]}]
+  (into #{} (comp (mapcat vals) cat) [base-names component-names]))
+
 (defn include-project? [{:keys [is-dev alias name]} selected-projects is-dev-user-input]
   (or (or (contains? selected-projects name)
           (contains? selected-projects alias))
@@ -55,6 +58,7 @@
         include-project (include-project? project selected-projects is-dev-user-input)
         project-has-changed (project-changed? project changed-projects)]
     (assoc project
+      ;; only ever includes bricks that have test sources
       :bricks-to-test
       (->> [:src :test]
            (into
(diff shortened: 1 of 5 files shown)


components/test/src/polylith/clj/core/test/bricks_to_test.clj
```diff
@@ -2,55 +2,87 @@
...
-                           ;; if we pass in :all or :all-bricks or if the project has changed then always run all brick tests.
-                           included-bricks
-                           (set/intersection included-bricks
-                                             selected-bricks
-                                             (into #{} cat
-                                                   [changed-bricks (source indirect-changes)])))
-                         #{})]
+        included-bricks (common/brick-names-to-test test eligible-bricks)
+        user-selected-bricks (if user-selected-bricks (set user-selected-bricks) eligible-bricks)
+        selected-bricks (if include-project
```
tengstrand: If the user doesn't add `brick:foo` or `project:bar` this function (bricks-to-test) will return (directly or indirectly) changed bricks, which we want to test. I think it's a bit misleading to change from `changed-bricks` to `selected-bricks` here, for that reason. 

I understand why you did the other change from `selected-bricks` to `user-selected-bricks`. But I would prefer to keep the name `selected-bricks`. The reason is that it's called that everywhere else in the call chain. We also have `selected-projects`. I think that can keep its name too, otherwise we should rename that to `user-selected-projects` too.

components/test/src/polylith/clj/core/test/bricks_to_test.clj
```diff
@@ -2,55 +2,87 @@
...
+                                      %
+                                      changed-bricks
+                                      user-selected-bricks
+                                      include-project
+                                      project-has-changed
+                                      is-run-all-brick-tests
+                                      (bricks-for-source project %))))
+           (vec))
+      ;; might include bricks that don't have test sources
+      :bricks-to-test-2
```
tengstrand: And yes, we need to come up with a better name. One suggestion is `:bricks-to-test-all-sources`.

tengstrand (changes requested):
I made a couple of comments, but except from that, I like your changes!

components/test/src/polylith/clj/core/test/bricks_to_test.clj
imrekoszo: I like it, thanks!

components/test/src/polylith/clj/core/test/bricks_to_test.clj
imrekoszo: Makes sense, I renamed them back.

tengstrand (approved):
Thanks for this!

imrekoszo:
Thank you!

Issue #533, opened 2025-03-01
Maps with tag literals as keys in test namespaces breaks test runner


EduardoMRB:
When there is a map with tag literals as keys on a test namespace, it throws an exception when `edamame` tries to parse it, and that results in an attempt to use an empty namespace in the test runner.

The culprit seems to be the addition of this line https://github.com/polyfy/polylith/blob/master/components/file/src/polylith/clj/core/file/core.clj#L138, as it makes tag literals to resolve to `nil`, such that:

```clojure
{#uuid "1e9183e0-26d7-44e0-967e-01d327c6ac10" 0,
 #uuid "0822ab64-6003-4f35-9113-8be279ac209a" 0.6,
 #uuid "460a65f9-988e-4045-b0f1-f4b2cd915ce2" 1}
```

becomes:
```clojure
{nil 0,
 nil 0.6,
 nil 1}
```

resulting in a `"Map literal contains duplicate key: "` when trying to parse.

Test error output:

```
Wrong number of args (0) passed to quote; clojure.lang.ExceptionInfo: Wrong number of args (0) passed to quote {:form (quote)} {:form (do (clojure.core/use (quote clojure.test)) (clojure.core/require (quote )) (clojure.test/run-tests (quote )))}
	at polylith.clj.core.test_runner_orchestrator.core$__GT_eval_in_project$fn__12818.invoke(core.clj:110)
	at polylith.clj.core.clojure_test_test_runner.core$run_test_statements$fn__36616.invoke(core.clj:59)
	at polylith.clj.core.clojure_test_test_runner.core$run_test_statements.invokeStatic(core.clj:58)
	at polylith.clj.core.clojure_test_test_runner.core$run_test_statements.invoke(core.clj:52)
	at polylith.clj.core.clojure_test_test_runner.core$create$reify__36642.run_tests(core.clj:103)
	at polylith.clj.core.test_runner_orchestrator.core$run_tests_for_project_with_test_runner.invokeStatic(core.clj:88)
	at polylith.clj.core.test_runner_orchestrator.core$run_tests_for_project_with_test_runner.invoke(core.clj:77)
	at polylith.clj.core.test_runner_orchestrator.core$run_tests_for_project.invokeStatic(core.clj:158)
	at polylith.clj.core.test_runner_orchestrator.core$run_tests_for_project.invoke(core.clj:128)
	at polylith.clj.core.test_runner_orchestrator.core$run.invokeStatic(core.clj:218)
[...]

PR #537, opened 2025-05-12
Maps with tag literals as keys (read correctly)

tengstrand:
Issue #533


seancorfield:
The fix for this is wrong -- and breaks on `#inst "2008"` (which is in our codebase at work and is valid Clojure!).

PR #560, opened 2025-08-21
docs(doc/project.adoc): uses correct plural `:projects` reference

johanmynhardt:
The correct key in `./workspace.edn` is `:projects`.


tengstrand (approved):
Thanks for fixing this typo!

PR #593, opened 2025-11-26
Fix npm scoped package handling in libs command

jxonas:
## Summary

  - Fix scoped npm packages (e.g., `@mantine/core`) showing incorrect latest versions in `poly libs :outdated`
  - Support npm/yarn workspaces when calculating package sizes

  ## Problem

  When using `poly libs :outdated`, npm packages with scoped names showed incorrect "latest" versions. For example:

  | library      | version  | latest  |
  |--------------|----------|---------|
  | charts       | ^8.3.8   | 0.0.3   |
  | core         | ^8.3.8   | 1.0.113 |

  The scope prefix was being stripped before querying the npm registry, causing lookups for unrelated packages (`core` instead of `@mantine/core`).

  Additionally, package size lookups failed for workspaces where dependencies are nested under the workspace package directory.

  ## Root Cause

  In Clojure, keywords containing `/` are interpreted as namespaced keywords. When `@mantine/core` from `package.json` becomes `:@mantine/core`:
  - `(namespace :@mantine/core)` → `"@mantine"`
  - `(name :@mantine/core)` → `"core"`

  Using `clojure.core/name` only returned the local name, dropping the scope.

  ## Solution

  1. Added `keyword->full-name` helper that preserves the namespace for scoped packages
  2. Updated npm registry queries to use full package names
  3. Added fallback logic for npm/yarn workspaces that checks both:
     - Nested: `node_modules/@poly/mantine/node_modules/@mantine/core`
     - Hoisted: `node_modules/@mantine/core`

  ## Files Changed

  - `components/antq/src/polylith/clj/core/antq/npm.clj`
  - `components/lib/src/polylith/clj/core/lib/size_npm.clj`
  - `components/lib/src/polylith/clj/core/lib/core.clj`

### DIFF
--- a/components/antq/src/polylith/clj/core/antq/npm.clj
+++ b/components/antq/src/polylith/clj/core/antq/npm.clj
@@ -2,13 +2,21 @@
   (:require [clj-http.client :as http]
             [cheshire.core :as json]))
 
+(defn keyword->full-name
+  "Convert a keyword to its full string name, preserving namespace for scoped npm packages.
+   For :@mantine/core, returns '@mantine/core' instead of just 'core'."
+  [kw]
+  (if-let [ns (namespace kw)]
+    (str ns "/" (name kw))
+    (name kw)))
+
 (defn npm-registry-url [package-name]
   (str "https://registry.npmjs.org/" package-name))
 
 (defn get-latest-version [package-name]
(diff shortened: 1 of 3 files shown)


tengstrand (approved):
Looks good! Thanks!

